PR #77, opened 2012-01-13
rpm package

thinker0:
rpm package hiredis.spec


badboy:
I am not a RedHat user nor do I have any experience with rpm packages. I will therefore not include this. It's better to rely on your distribution's package maintainer for this anyway.

PR #129, opened 2012-10-28
Optional pkg-config and CMake configuration files

thomaslee:
Optional pkg-config and CMake configuration files.
    
With these changes, you can now automatically retrieve hiredis include & library info from pkg-config (e.g. pkg-config hiredis --cflags --libs).
    
A CMake configuration file is also provided, which makes it easy for folks to link against hiredis in projects managed by CMake (e.g find_package(Hiredis))

The CMake support is currently built on top of the pkg-config stuff, and so will only be installed if we manage to detect the appropriate location for the .pc file, but no reason why it couldn't work independently with a little more effort.

These features are completely optional & the build will automatically detect at install time whether the appropriate pkg-config / CMake directories are in place before attempting to install the configuration files.


Qix-:
@pietern adding CMakeLists.txt allows consuming applications to simply `add_subdirectory("path/to/hiredis")` and then `target_link_libraries(my-project hiredis)` to wire everything up.

While I can see why CMake's language is annoying, CMake-enabled applications are very clearly spec'd and have a very clear path to incorporation. They also allow other build systems to be used - for example, we prefer Ninja as it parallelizes builds a bit better. CMake supports it out of the box.

ChrisCates:
What I actually end up doing is just this:

```cmake
IF (NOT HIREDIS_LIBRARIES)
  MESSAGE("-- Setting default Hiredis Library Directory")
  SET(HIREDIS_LIBRARIES "/usr/local/lib/libhiredis.a")
ENDIF (NOT HIREDIS_LIBRARIES)

MESSAGE("-- HIREDIS_LIBRARIES\t" ${HIREDIS_LIBRARIES})

IF (NOT HIREDIS_INCLUDE_DIRS)
  MESSAGE("-- Setting default Hiredis Include Directory")
  SET(HIREDIS_INCLUDE_DIRS "/usr/local/include/hiredis")
ENDIF (NOT HIREDIS_INCLUDE_DIRS)

MESSAGE("-- HIREDIS_INCLUDE_DIRS\t" ${HIREDIS_INCLUDE_DIRS})

INCLUDE_DIRECTORIES(${HIREDIS_INCLUDE_DIRS})
LINK_DIRECTORIES(${HIREDIS_INCLUDE_DIRS})

SET(
  REQUIRED_LIBRARIES
  ${REQUIRED_LIBRARIES}
  ${HIREDIS_LIBRARIES}
)
```

Hopefully this is helpful for other people using cmake. 💯

PR #150, opened 2013-01-27
Memory footprint optimization

hdoreau:
Updated tests and documentation accordingly.

### DIFF
--- a/async.c
+++ b/async.c
@@ -347,9 +347,11 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
     /* Custom reply functions are not supported for pub/sub. This will fail
      * very hard when they are used... */
     if (reply->type == REDIS_REPLY_ARRAY) {
-        assert(reply->elements >= 2);
-        assert(reply->element[0]->type == REDIS_REPLY_STRING);
-        stype = reply->element[0]->str;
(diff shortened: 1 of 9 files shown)


brendanashworth:
Assuming hiredis is on semver, it is still on major version 0 so a major bump wouldn't be explicitly necessary. However, if you're still on major version 0 after this long, there should have been a 1.0.0 release a long time ago.

async.c
```diff
@@ -347,9 +347,11 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
     /* Custom reply functions are not supported for pub/sub. This will fail
      * very hard when they are used... */
     if (reply->type == REDIS_REPLY_ARRAY) {
-        assert(reply->elements >= 2);
-        assert(reply->element[0]->type == REDIS_REPLY_STRING);
-        stype = reply->element[0]->str;
+        redisReply *elm = reply->reply_array.element[0];
```
wichmanndell: would you be better of accessing the reply->reply_array this after the assert?

badboy:
I'm not pulling this in. Both because hiredis is used upstream in hiredis and I doubt it is worth the change there and because, as pietern said, the API get's more verbose. The memory optimization seems minor in this case.

PR #204, opened 2013-11-30
fix type conversion in libuv.h

chishaxie:


mattsta:
Fixed in f20f5ad10264a10d41aa2a47c029fba71820d62a

PR #230, opened 2014-04-07
Remove possiblity of multiple close on same fd

mattsta:
With all the async connects and disconnects and error handling
going on in hiredis, we need to centralize how we close our fd
and set it so it doesn't get re-closed.  Prior to this commit,
sometimes we'd close(fd), run an async error handler, then
call close(fd) again.

To stop multiple closes, we now set fd to -1 after we close it,
but that requires not passing fd as an independent argument to
functions.

This commit moves all fd usage to c->fd.  Since the context
has a fd field and all functions receive the context, it makes
more sense to use the fd inside of c instead of passing along fd
as an independent argument.

Also, by only using c->fd, we can set c->fd after we close it to
signify we shouldn't re-close the same fd again.

This does change one semi-public interface function redisCheckSocketError()
to only take (context) instead of (context, fd).  A search on github
returned zero occasions of people using redisCheckSocketError()
outside of net.{c,h} in hiredis itself.

Commit inspired by the bug report at:
https://groups.google.com/forum/#!topic/redis-db/mQm46XkIPOY

Thanks go out to Thijs for trying high-frequency reconnects on
a host that isn't there.

All tests pass.

### DIFF
--- a/net.c
+++ b/net.c
@@ -226,14 +226,12 @@ int redisCheckSocketError(redisContext *c) {
 
     if (getsockopt(c->fd, SOL_SOCKET, SO_ERROR, &err, &errlen) == -1) {
         __redisSetErrorFromErrno(c,REDIS_ERR_IO,"getsockopt(SO_ERROR)");
-        redisContextCloseFd(c);
         return REDIS_ERR;
     }
 
     if (err) {
         errno = err;
         __redisSetErrorFromErrno(c,REDIS_ERR_IO,NULL);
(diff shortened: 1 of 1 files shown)


GeekCome:
I am using a version before this patch. But the program is just one thread. The error case just like this bug. The event handled by libevent, and the epoll_wait returns a active event, but libevent just leave there without handling.
The issue I commit https://github.com/libevent/libevent/issues/230

PR #258, opened 2014-07-22
Static declaration of 'redisLibuvAttach'

truongminh:
Fix static declaration of 'redisLibuvAttach' follows non-static declaration


mattsta:
Fixing this issue using #251 instead.  Thanks for the suggestion!

PR #261, opened 2014-08-04
Increase maximum SDS size

mattsta:
SDS sizes can never be negative, so it's safe to double their
capacity from 2 GB to 4 GB.

This helps extend some edge cases where people request comically large results from Redis.  Better to handle as much data as we can without running out of space.


badboy:
Superseeded by sds v2 as provided in #347

PR #284, opened 2014-12-06
Test hiredis bindings

thedrow:
[Review on Reviewable](https://reviewable.io/reviews/redis/hiredis/284)


badboy:
Only works for me with:

```
git submodule init && \
  git submodule update && \
  git submodule foreach git pull origin HEAD && \
  git submodule status
```

(tested locally)

Also, if this should be considered for merge please squash the commits down to one.

badboy:
Another problem I see: these tests will only check against latest master, but _not_ PRs or similar, so all in all the will miss most problems.

PR #297, opened 2015-01-23
Makefile fixes

wiedi:
This PR consists of three changes to the Makefile:

1.) added support for the DESTDIR variable, this helps with packaging - if not used nothing changes
2.) fix INSTALL_PKGCONF_PATH to the correct path which should depend on INSTALL_LIBRARY_PATH
3.) since hiredis.h now includes both read.h and sds.h the install target should also install those.


badboy:
We should pull this in and make 0.12.1 I guess.

mattsta:
Thanks for the solutions! Sorry for the initial breakage. This commit is the only major user-facing change in 0.12.1.

PR #302, opened 2015-01-30
Fix make install again

cicku:
For real, the commit 4355ab3c8fa7ad98b8a21267f84ebbaf47e2d2f8 messed up with me here. At least I couldn't build 0.12.0+ for Fedora without harsh patch.

It installed files to somewhere outside first, and then destdir was introduced but abused. Meanwhile the pkgconfig file was broken, the original echo line in makefile didn't escape the dollar sign so the final generated file would ignore the ${includedir} and ${libdir} as they were empty. I know modern compiler is smart but we still need some pedantic styles as they are not matured yet on such style it used.

This patch:

1. Use $(DESTDIR) only in install stage.
2. Change pkgconfig file from:

Libs: -L -lhiredis
Cflags: -I -D_FILE_OFFSET_BITS=64

to

Libs: -L${libdir} -lhiredis
Cflags: -I${includedir} -D_FILE_OFFSET_BITS=64

3. $(PKGCONFNAME) stage is needed for install stage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,8 @@ PREFIX?=/usr/local
 INCLUDE_PATH?=include/hiredis
 LIBRARY_PATH?=lib
 PKGCONF_PATH?=pkgconfig
-INSTALL_INCLUDE_PATH= $(DESTDIR)$(PREFIX)/$(INCLUDE_PATH)
-INSTALL_LIBRARY_PATH= $(DESTDIR)$(PREFIX)/$(LIBRARY_PATH)
+INSTALL_INCLUDE_PATH= $(PREFIX)/$(INCLUDE_PATH)
+INSTALL_LIBRARY_PATH= $(PREFIX)/$(LIBRARY_PATH)
 INSTALL_PKGCONF_PATH= $(INSTALL_LIBRARY_PATH)/$(PKGCONF_PATH)
 
 # redis-server configuration used for testing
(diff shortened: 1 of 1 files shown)


mattsta:
So, who's right?  This removes a lot of https://github.com/redis/hiredis/commit/4355ab3c8fa7ad98b8a21267f84ebbaf47e2d2f8 for fixing installs around packaging.

(The build isn't really failing; ubuntu mirrors are just down and travis can't complete the VM install.)

cicku:
I think mine is better. 

The initial way of fixing the make install problem is a harsh way. It added DESTDIR to INSTALL_LIBRARY_PATH and the pkgconfig file would be broken because the actual libdir path is /usr/lib or /usr/lib64, but the fix would pull in the DESTDIR macro. This also applies to the INSTALL_INCLUDE_PATH.

DESTDIR is a good macro as I don't need to insert it manually after 0.12.0, but it should be used in a proper way, the commit needs to fit more system or the troubles will come more.  

Do we need to toss a coin?  😁

mattsta:
> Do we need to toss a coin? 😁

Look how [other Makefiles implement DESTDIR](https://github.com/search?p=1&q=DESTDIR+filename%3AMakefile&ref=searchresults&type=Code&utf8=%E2%9C%93) and see if we are right or wrong?  It looks like we are currently doing the right thing.

cicku:
Mine: Include the DESTDIR explicitly(in INSTALL_*_PATH macros)

Initial: Implicitly(directly in the install line)

To be frank, DESTDIR is not a big problem, I'm concerned about the pkgconfig file.

I don't have too much time arguing on DESTDIR as I'm tired of such kind of minutia. It's not a real problem indeed. No matter how you cope with this finally, I will carry this patch in Fedora package forever and will make sure hiredis could be built without issue and make sure the pkgconfig is correct instead of a broken one included in the RPM.

Thank you in advance.

wiedi:
Couldn't the DESTDIR for pkgconfig fixed by just changing lines 154, 155 to:

```
 	@echo libdir=$(PREFIX)/$(INCLUDE_PATH) >> $@
 	@echo includedir=$(PREFIX)/$(LIBRARY_PATH) >> $@
```

Either way, from looking at it nothing changes in regards to where stuff gets installed so LGTM. (Didn't test it yet)

Sorry that 4355ab3c8fa7ad98b8a21267f84ebbaf47e2d2f8 wasted your time and thanks for catching the pkgconfig issues.

mattsta:
> Couldn't the DESTDIR for pkgconfig fixed by

Good point.  The previous change prepended DESTDIR to the vars we were using for pkgconf, so we can easily make pkgconf use the previous values directly.

I'll include your fix soon.

cicku:
Thank you everyone.

PR #311, opened 2015-02-28
Fix hiredis.h include path in example.c

krisis:
Without this change, examples/example.c fails to compile.
```
[kp@trantor examples (master)]$ gcc -lhiredis example.c 
example.c:5:21: fatal error: hiredis.h: No such file or directory
 #include <hiredis.h>
                     ^
compilation terminated.
```
Signed-off-by: Krishnan Parthasarathi <kp@gluster.org>

### DIFF
--- a/examples/example.c
+++ b/examples/example.c
@@ -2,7 +2,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#include <hiredis.h>
+#include <hiredis/hiredis.h>
 
 int main(int argc, char **argv) {
     unsigned int j;


badboy:
Except that breaks our own Makefile:
r Makefile:

```
% make hiredis-example             
cc -o examples/hiredis-example -O3 -fPIC  -Wall -W -Wstrict-prototypes -Wwrite-strings -g -ggdb    -I. examples/example.c libhiredis.a
examples/example.c:5:29: fatal error: hiredis/hiredis.h: No such file or directory
 #include <hiredis/hiredis.h>

                             ^
compilation terminated.
Makefile:117: recipe for target 'hiredis-example' failed
make: *** [hiredis-example] Error 1
```

For an external program using hiredis as a library `hiredis/hiredis.h` is the right thing (we install files into `.../include/hiredis/`).
We also use `include <hiredis.h>` in all other examples. I'd say we just stay with that for now and include documentation about how to include in the Readme.

krisis:
OK. That makes sense. I shall close this pull request.

PR #312, opened 2015-03-03
Fix pkgconf file: escaping needed

DimStar77:
Due to the various processors going over the command, we need more
escaping.

1) Make parses it, so $${libdir} becomes ${libdir}
2) 'shell' parses it for the 'echo command', whereas echo ${libdir}
would be an empty string; escape it as \${libdir} to ensure we get what
we want.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,8 +157,8 @@ $(PKGCONFNAME): hiredis.h
 	@echo Name: hiredis >> $@
 	@echo Description: Minimalistic C client library for Redis. >> $@
 	@echo Version: $(HIREDIS_MAJOR).$(HIREDIS_MINOR).$(HIREDIS_PATCH) >> $@
-	@echo Libs: -L$${libdir} -lhiredis >> $@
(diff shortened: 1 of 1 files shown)


mattsta:
Which platform or version of make has the problem?

DimStar77:
I'm running all my tests on openSUSE Tumbleweed, currently make 4.1

without this patch, the resulting .pc file looks like this:

``prefix=/usr
exec_prefix=
libdir=/usr/lib64
includedir=/usr/include/hiredis

Name: hiredis
Description: Minimalistic C client library for Redis.
Version: 0.12.1
Libs: -L -lhiredis
Cflags: -I -D_FILE_OFFSET_BITS=64
```
Note the invalid -L in Libs: it must be followed by a path, as well as -I in Cflags.

mattsta:
Oh, you're right.  Obviously wrong before.  Sorry for the broken file and thanks for the quick fix.

PR #314, opened 2015-03-13
Added support for compiling the parser code with Microsoft Visual C compiler.

tzickel:
For hiredis-py and others support on windows.

This patch is the final piece in the refactoring of hiredis and hiredis-py (and any other lib which only needs the parser code of hiredis) to make it work on windows

### DIFF
--- a/read.c
+++ b/read.c
@@ -33,7 +33,9 @@
 #include "fmacros.h"
 #include <string.h>
 #include <stdlib.h>
+#ifndef _MSC_VER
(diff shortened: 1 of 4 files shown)


badboy:
Travis tests broken for some weird network-related reason. Local tests on Linux still work fine.
And the best thing: looks like hiredis-py compiles on Windows with this changes as well: https://ci.appveyor.com/project/badboy/hiredis-py-win

PR #316, opened 2015-03-19
Fix hiredis.pc generation.

boardwalk:
INCLUDE_PATH and LIBRARY_PATH were swapped!


badboy:
Thanks. I'll make sure we don't break the `hiredis.pc` file again.

PR #318, opened 2015-03-28
Add PKGCONFNAME to install dependencies.

neonquill:
Attempting to use the install target before the make target works fine,
except for the missing pkgconfig file.  Adding that file to the
dependencies for the install target to make sure it gets created first.


badboy:
Thanks! That was indeed missing.

PR #319, opened 2015-03-31
Request timeouts. Fixes #306

### LINKED ISSUE #306: Add timeout option to requests
While connectWithTimeout will guard against the use case of a slow initial handshake, there's nothing that guards against delays with requests on current connections. The example case here is if Redis spikes to 100% CPU utilization for any number of reasons (out of sockets, memory issues, etc) then anything left holding a connection will suffer without the ability to timeout. I have run into this a couple times now and would like to see it solved. I am happy to implement, but looking for at least a thumbs up on interest before I start.

abedra:
This adds support for timing out requests. It addresses issue #306.

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -658,7 +658,9 @@ void redisReconnect(redisContext *c) {
     c->obuf = sdsempty();
     c->reader = redisReaderCreate();
 
-    if (c->ip != NULL) {
+    if (c->source_addr != NULL && c->ip != NULL) {
+	redisContextConnectBindTcp(c,c->ip,c->port,NULL,c->source_addr);
+    } else if (c->ip != NULL) {
 	if (c->timeout == NULL) {
 	    redisContextConnectTcp(c, c->ip, c->port, NULL);
 	} else {
@@ -670,8 +672,6 @@ void redisReconnect(redisContext *c) {
 	} else {
 	    redisContextConnectUnix(c, c->path, c->timeout);
 	}
-    } else if (c->source_addr != NULL && c->ip != NULL) {
-	redisContextConnectBindTcp(c,c->ip,c->port,NULL,c->source_addr);
     } else {
 	/* Something bad happened here and shouldn't have. There isn't
 	   enough information in the context to reconnect. */
--- a/hiredis.c
+++ b/hiredis.c
@@ -599,6 +599,8 @@ static redisContext *redisContextInit(void) {
     c->errstr[0] = '\0';
     c->obuf = sdsempty();
     c->reader = redisReaderCreate();
+    c->ip = sdsempty();
+    c->port = 6379;
 
     if (c->obuf == NULL || c->reader == NULL) {
         redisFree(c);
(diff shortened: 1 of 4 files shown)


badboy:
Would be easier if your code wouldn't change the indentation (then it would also not touch unchanged code).

If we implement a `reconnect` method, we also need to implement this for unix sockets, thus we also need to track the path (and track whether it was TCP or unix socket in the first place).
We can also just move the assignment of `ip` and `port` into `redisContextConnectTcp` instead of duplicating it for each tcp connection method.

I will take it from here tomorrow. thanks for this. :)

abedra:
Ok, that seems reasonable. What is the next step on this then?

abedra:
Bump. What is the next step here? I can move this forward this week if you don't have the time.

badboy:
* Provide a indentation-correct PR
* Implement reconnect for Unix sockets as well
* We got some more TCP connection methods, not sure if we need to save state for these as well 
* Check if we need more tests for this. I guess reconnect for TCP and unix should be fine.

You can base your work against the [with-timeout](https://github.com/redis/hiredis/commits/with-timeout) branch, which already has the force-error-on-timeout and thus only needs the reconnect.

abedra:
Ok, I added the functionality to the remaining functions and cleaned up the indentation issues.

badboy:
Thanks, from the first look this seems good, I'll check properly later and will try to get this finished this week.

abedra:
Was reviewing further and found a small issue. One more commit in the pipeline to fix that.

abedra:
Anything left to do here or can this be merged?

badboy:
Likely yes, it's on my todo for tomorrow. Sorry for the long wait, busy week at university.

mezzatto:
Whats wrong with redisSetTimeout()? I've been using it for years without problems...
It sets the timeout for ongoing connections (the data transfer timeout). Woulds these commandWithTimeout() be messing with that? What happens if I do redisSetTimeout(10) but commandWithTimeout(1) or commandWithTimeout(100)? Which one will prevail?

badboy:
Indeed we have that, I completely overread it in #306, but it wasn't clarified any further by @abedra why we need the new `withTimeout` command.
I changed the test to use `redisSetTimeout` instead and it passes. @abedra: Any followup on what's the problem with `redisSetTimeout` then?

We can still benefit from the reconnection methods, right?

abedra:
The only reason I did this was because that didn't seem to work for me. I suppose it's possible that it was fixed at some point past the version I was using, but I also asked the initial question in #306 and it was suggested that I go ahead with this work.

mezzatto:
@abedra, I suppose that it makes sense if you use redisSetTimeout() with some high value (say 100) but, for some reason, want a smaller value (say 10) to a specific command using the same connection. But it looks confusing to me. Users will create bugs in their software with this behavior...

mezzatto:
I think the way to go if we really want the ```withTimeout``` command is to do something that ignores the previous timeout set by ```redisSetTimeout()``` so that if you do ```redisSetTimeout(10)``` and then ```commandWithTimeout(100)``` it will wait for 100 instead of 10.

I think that is possible if you backup the current timeout of the socket, set it to 0, do the ```withTimeout``` and then restore the backed up value.

But that looks ugly to me :(

badboy:
@abedra If you can reconstruct a case where it doesn't work as intended that would be great. I'm very sorry for the lost time for working on this, neither me nor @mattsta fully checked why `redisSetTimeout` wouldn't work and encouraged you. I'm sorry for that

@mezzatto I don't really see the need anymore. If `SO_RCVTIMEO` works as expected we can skip all the hassle with our new commands. If a user wants a longer timeout for certain commands, he can just reset the timeout before issuing the command. No need for `hiredis` providing yet another API for this.

abedra:
@badboy I'm extremely disappointed here. Not only have I had a production bug since #306 was opened, we have obviously had a lot of back and forth with more than one maintainer. The wasted code doesn't bother me a bit, but the wasted time does. If this was really the conclusion the entire time why didn't the maintainers of this project simply say so. I'm still not convinced it works properly, but the situation that came up is almost impossible to reproduce at this point for me. 

Regardless of this, the reconnect logic is still going to be necessary since the timeout sets an error. If I send a new pull request with only the reconnect code, will you merge it in a timely fashion?

badboy:
I fully understand your disappointment and again I'm deeply sorry it happened this way. Neither me nor @mattsta wrote this project in the first place, we both took over maintainership only this year, with me doing this only in my free time. 
When first taking over this issue I was thinking that Matt already thought it through. It now shows that neither me nor Matt has the full understanding of the code yet, so this was overlooked. I'm deeply sorry for the wasted time.

I will merge the reconnect code ASAP, probably tonight. I just cherry-pick it from this PR, no need for another PR from your side.

abedra:
Ok thanks. I'll watch for the merge. Will you be tagging a new release after this?

badboy:
Yep, right after this I'll tag a new release

PR #322, opened 2015-04-28
Initial CMake build system integration

jerryjacobs:
This is a initial working draft of the CMake build system building and running hiredis test.
Hope you like it.


badboy:
I'm not really a fan of cmake nor do I have any clue how it works, so I won't include it here, our Makefile works just fine.

PR #325, opened 2015-04-29
avoid name clash with system specific predefined macro 'unix'

### LINKED ISSUE #321: issue in hiredis.h
Hi, I am using hiredis.h in my program, and when I compile my program an error return "/usr/local/include/hiredis/hiredis.h:158: error：expected unqualified-id before numeric constant". I checked that the line of 158 in hiredis.h is a struct :
struct {
        char *path;
    } unix;//this is line 158

Is there any complicit in centos when using hiredis.h?

ncopa:
It is common to find 'unix' defined on Unix systems before c99. While we
can force that hiredis itself is compiled with c99, we cannot really
enforce that all projects using the hiredis headers are compiled with
c99.

So avoid the use of 'unix' in the public headers.

The problem was introduced with commit d9e0b0f6 (Implement a reconnect
method for the client context, 2015-04-16)

See also:
http://stackoverflow.com/questions/3770322/is-unix-restricted-keyword-in-c

This fixes #321

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -600,9 +600,9 @@ static redisContext *redisContextInit(void) {
     c->errstr[0] = '\0';
     c->obuf = sdsempty();
     c->reader = redisReaderCreate();
-    c->tcp.host = NULL;
-    c->tcp.source_addr = NULL;
-    c->unix.path = NULL;
+    c->c_tcp.host = NULL;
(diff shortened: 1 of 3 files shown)


badboy:
I'm pulling in #326 for the simpler rename. Thanks for this anyway!

PR #330, opened 2015-05-26
redisBufferRead: Clear REDIS_CONNECTED flag when connection is closed

xor-gate:
The redisContext flag REDIS_CONNECT is not cleared when REDIS_ERR_EOF is returned in redisBufferRead. For public API the c->err = REDIS_ERR_EOF is not known. Clearing the REDIS_CONNECT flag is a good practice.

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -808,6 +808,7 @@ int redisBufferRead(redisContext *c) {
             return REDIS_ERR;
         }
     } else if (nread == 0) {
+        c->flags &= ~REDIS_CONNECTED;
         __redisSetError(c,REDIS_ERR_EOF,"Server closed the connection");
         return REDIS_ERR;
     } else {


rikvdh:
+1

Clearing the REDIS_CONNECT makes it clear to the user that the connection terminated, it prevents users from excessive reconnecting.

badboy:
I understand why it makes sense to clear the connect flag, but why do you say that `REDIS_ERR_EOF` is not known in the public API?

hiredis.c
```diff
@@ -808,6 +808,7 @@ int redisBufferRead(redisContext *c) {
             return REDIS_ERR;
         }
     } else if (nread == 0) {
+        c->flags =~ REDIS_CONNECTED;
```
badboy: This should probably be `c->flags &= ~REDIS_CONNECTED;`, right?

xor-gate: Yes small typo.

xor-gate:
`REDIS_ERR_EOF` is not known in https://github.com/redis/hiredis/blob/master/hiredis.h only there is a  `err` field in `struct redisContext`

xor-gate:
I fixed it in my branch with some help of @rikvdh :-)

badboy:
`REDIS_ERR_EOF` is defined in [read.h](https://github.com/redis/hiredis/blob/master/read.h#L45), and, exactly because it is an error state and thus visible to the user, also available as part of the public API (`read.h` is included in `hiredis.h`)

xor-gate:
I overlooked this and for me a more convenient place would be hiredis.h

badboy:
:) No problem, I know the documentation is currently lacking a bit.

rikvdh:
@badboy Can this be merged? I think this is still a valuable addition (apart from the lacking docs)

PR #347, opened 2015-07-24
apply sds from redis

charsyam:
After patching sdshdr size optimization.
Sentinel is broken. 
because sentinel uses hiredis and it's sds is old(just fixed size sdshdr)
so, we should apply new sdshdr into hiredis also.

### DIFF
--- a/sds.c
+++ b/sds.c
@@ -1,6 +1,6 @@
-/* SDS (Simple Dynamic Strings), A C dynamic strings library.
+/* SDSLib, A C dynamic strings library
  *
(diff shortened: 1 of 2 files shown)


badboy:
Absolutely. I will first release a bug fix release next week and then will put this in the next feature release.

charsyam:
@badboy before merging. please check https://github.com/antirez/redis/pull/2696 this discussion :)
Thanks :)

badboy:
Included in #410

PR #349, opened 2015-07-31
Added broken connection handling to libuv adapter

sopindm:
Libuv adapter hangs out when connection fails. This fix adds fake write event that handles this case

### DIFF
--- a/adapters/libuv.h
+++ b/adapters/libuv.h
@@ -7,16 +7,16 @@
 #include <string.h>
 
 typedef struct redisLibuvEvents {
-  redisAsyncContext* context;
-  uv_poll_t          handle;
-  int                events;
+  redisAsyncContext *context;
+  uv_poll_t handle;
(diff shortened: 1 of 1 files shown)


adapters/libuv.h
```diff
@@ -12,11 +12,11 @@ typedef struct redisLibuvEvents {
   int                events;
 } redisLibuvEvents;
 
-
 static void redisLibuvPoll(uv_poll_t* handle, int status, int events) {
   redisLibuvEvents* p = (redisLibuvEvents*)handle->data;
 
   if (status != 0) {
+    redisAsyncHandleWrite(p->context);
```
wangdaojian: test

sopindm:
Error handling is fixed in 'unstable' branch. But issue with getting double (EV_READ and EV_WRITE) events on close is still up, isn't it?

PR #354, opened 2015-08-23
Moved to containerized infrastrucutre and activated ccache

thedrow:


thedrow:
It seems that one of the required packages are not whitelisted. Will fix.

thedrow:
It seems that @badboy has some of my changes in unstable.
However I enabled ccache and used colored output for make.
I can make a new pull request with those against the unstable branch if needed.
What do you guys think?

badboy:
Not a big fan of the colored make output.
Need to check whether ccache is worth it, but it's simple to add, so I might try it.
The rest is indeed already handled in the unstable branch.

PR #356, opened 2015-08-31
Add libevent2 adapter

tattsun:
No adapter for libevent2 doesn't seems to be available.(and I'm using libevent2 now :cry:)
How about supporting libevent2?


badboy:
The current libevent adapter seems to work with libevent 2.0.22-1
Isn't that libevent2 already? What's different about your adapter?

tattsun:
oh, I've misunderstood.
sorry..... :sob:

PR #357, opened 2015-09-01
(async) connect and disconnect callbacks should receive privData

tattsun:
In my opinion, async connect and disconnect callbacks should be able to receive privData for C++ users and so on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 /*.dylib
 /*.a
 /*.pc
+CMakeLists.txt
+.idea/*
--- a/async.c
+++ b/async.c
@@ -41,6 +41,16 @@
 #include "dict.c"
 #include "sds.h"
 
+struct redisConnectCallbackPrivObj {
+    redisConnectCallbackWithPrivdata *fn;
+    void *privData;
+};
+
+struct redisDisconnectCallbackPrivObj {
+    redisDisconnectCallbackWithPrivdata *fn;
+    void *privData;
+};
+
 #define _EL_ADD_READ(ctx) do { \
         if ((ctx)->ev.addRead) (ctx)->ev.addRead((ctx)->ev.data); \
     } while(0)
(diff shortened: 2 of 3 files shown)


echoma:
In your commit, it seems that the privData is binded to the context, not to the connect() and disconnect(). This is different from redisvAsyncCommand() which bind the privData with this function calling. Users can maintain the privData by theirselves during the whole connection eastablishing period very easily.

tattsun:
You mean that we should hide privData from users..?

How about this change?

echoma:
'''
void* priv_data = malloc(sizeof(some_data_type))
void on_connect(const redisAsyncContext*, int)
{
    printf("this is my priv data %p", priv_data);
}
redisAsyncSetConnectCallback(context, on_connect);
'''
As in the above code, users can maitain the priv data very easily. The libhiredis don't need to care about this.

tattsun:
It's true when there is only one connection.

```
char* target_hosts[3] = {
  "redis1", "redis2", "redis3"
};
void on_connect(const redisAsyncContext*, int, void* host)
{
  printf("connected to %s\n", host);
}
redisAsyncSetConnectCallbackWithPrivdata(context1, on_connect, target_hosts[0]);
redisAsyncSetConnectCallbackWithPrivdata(context2, on_connect, target_hosts[1]);
redisAsyncSetConnectCallbackWithPrivdata(context3, on_connect, target_hosts[2]);
```

In the above example, redisAsyncSetConnectCallbackWithPrivdata are near each.
But, in the real world, redisAsyncSetConnectionCallback would be called at various places.

In my opinion, it's convenient that each callback could receive privdata in the real world.

And.. we will be able to easily implement C++ wrapper(each instance should receive connect callback).

echoma:
while not using redisAsyncContext's data member?
```
class Wrapper
{
public:
    redisAsyncContext* ctx_;
    void* priv_data_;

    void start_connect()
    {
        ctx_->data = this;
        redisAsyncSetConnectCallback(ctx_, Wrapper::on_connect);
    }
    static void on_connect(const redisAsyncContext* c, int)
    {
        Wrapper* w = (Wrapper*)(c->data);
        printf("this is my priv data %p", w->priv_data_);
    }
}
```

tattsun:
I didn't know that member..
Thank you and sorry for my misunderstanding.

PR #377, opened 2015-11-11
Add redisFormattedCommand() as an option for hiredis sync calls.

danielhtshih:
Sometimes we need to pre-format the incoming command for decision making, before we actually make the hiredis command call.

Since we already have redisAsyncFormattedCommand() for hiredis async calls; it is also useful to have redisFormattedCommand() for hiredis sync calls.

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -1019,3 +1019,9 @@ void *redisCommandArgv(redisContext *c, int argc, const char **argv, const size_
         return NULL;
     return __redisBlockForReply(c);
 }
+
+void *redisFormattedCommand(redisContext *c, const char *cmd, size_t len) {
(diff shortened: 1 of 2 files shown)


badboy:
You can already achieve the smae by using `redisAppendFormattedCommand` followed by a `redisGetReply` call, right?

danielhtshih:
Yes, as the redisCommand* family of functions, they mostly are the combination of ```redisCommand*``` and ```redisGetReply``` (called by ```__redisBlockForReply```).

One of the advantage of using ```redisFormattedCommand``` rather than ```redisAppendFormattedCommand``` followed by a ```redisGetReply``` is it can leverage the condition check in ```__redisBlockForReply```, which isn't intended to be directly called externally.

PR #379, opened 2015-11-18
Fix potential race in 'invalid timeout' tests

thomaslee:
It's possible for the call to connect() to succeed on the very first try, in which case the logic for checking for invalid timeout fields is never executed. When this happens, the tests fail because they expect a REDIS_ERR_IO but no such failure has occurred.

Tests aside, this is a potential source of irritating and hard-to-find intermittent bugs.

This patch forces the validation to occur early so that we get predictable behavior whenever an invalid timeout is specified.


badboy:
@not-a-robot r+

Issue #388, opened 2016-01-07
Gracefully closing down SUBSCRIBED clients


rikvdh:
I'm building a multithreaded application which reads events from Redis using hiredis. I'm using PSUBSCRIBE to subscribe.

Then, I use the following code (simplified) to read from Redis. Upon shutdown I change the stopping to true and it should stop. But it never reaches the code because hiredis is blocked in redisGetReply.

``` c
while (!stopping) {
    redisret = redisGetReply(ctx, &reply);
    if (redisret == REDIS_OK) {
        /* do stuff */
    }
}
```

As seen below in the partial backtrace it is stuck at read(). As far as I know hiredis has no `redisGetReplyTimeout` which uses select/poll for applications which must be able to shutdown gracefully.

GDB:

```
#0  0x00007ffff7bcdadd in read () at ../sysdeps/unix/syscall-template.S:81
#1  0x0000555555aca56b in redisBufferRead ()
#2  0x0000555555acaa16 in redisGetReply ()
```

Can this be implemented?

Thanks,

Rik

badboy:
You should be able to do one of

1. Set the context to non-blocking (`redisConnectNonBlock`) and handle the polling yourself
2. Use the async part of hiredis (with your prefered async lib)
3. Get the file descriptor out of the `redisContext` and do the select yourself and only call `redisGetReply` if it says you can read.

rikvdh:
@badboy Shouldn't option 3 get part of hiredis? I can make a PR for it if you like? Having users rumble around with fd's of the context shouldn't be the way to go. Perhaps a `redisSelect` or `redisSelectReply` and then combined with `redisGetReply` create a `redisGetReplyTimeout` function could be a great addition.

What do you think?

xor-gate:
No response @badboy ? Probably others will benefit to this to have a abstraction between the application and the library.

Asmod4n:
Can't you just poll() on the redisContext->fd ?

badboy:
Yes, poll on the file descriptor. At this point I will not add a specific method for this.

PR #389, opened 2016-01-08
add function to check if the redis-socket has data incoming

rikvdh:
See issue #388


xor-gate:
Looks good, maybe there should be a unit test?

badboy:
Given the whole content of the `redisContext` object is public, this can be implemented easily on the user side.

Code-wise I don't see the point in always returning `REDIS_OK` and using an out parameter for the actual value.

For now I'm not gonna pull this in.

PR #390, opened 2016-01-12
Try to prevent integer overflow in sds length.

cjmay:
I encountered integer overflows in sds when using hiredis to send many commands in a pipeline.  These manifested as hard-to-diagnose errors in my application.  I think it would be preferable to try to prevent overflows and fail with an informative error message.  This change affects the sds API substantially; I'm open to alternatives.

Commit message: try to maintain the invariant that
len + free is never larger than the maximum value
of the integer type.  Add new status flag to
several sds commands, setting it to SDS_OK by
default and SDS_ERR_LEN when there is an attempt
to violate this invariant.

### DIFF
--- a/async.c
+++ b/async.c
@@ -368,6 +368,7 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
     int pvariant;
     char *stype;
     sds sname;
+    int status = SDS_OK;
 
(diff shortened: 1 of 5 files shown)


badboy:
I will pull in the latest sds from Redis itself in the coming weeks, which should probably help with that.

cjmay:
Thank you!

PR #391, opened 2016-01-16
Prevent buffer overflow when formatting the error

### LINKED ISSUE #380: __redis_strerror_r pointer cross-border when define _GNU_SOURCE macro
version: 0.13
file:hiredis.h

if use GNU strerror_r, the strerror_r return static buffer,  then buf[(len)] will overwrite the redisContext pointer, lead to fd equal zero.
the original code:

```C
/* "bad" GNU strerror_r we need to clean up after. */
#define __redis_strerror_r(errno, buf, len)                                    \
    do {                                                                       \
        char *err_str = strerror_r((errno), (buf), (len));                     \
        /* If return value _isn't_ the start of the buffer we passed in,       \
         * then GNU strerror_r returned an internal static buffer and we       \
         * need to copy the result into our private buffer. */                 \
        if (err_str != (buf)) {                                                \
            buf[(len)] = '\0';                                               \
            strncat((buf), err_str, ((len) - 1));                              \
        }                                                                      \
    } while (0)
```
fixed:

```C
/* "bad" GNU strerror_r we need to clean up after. */
#define __redis_strerror_r(errno, buf, len)                                    \
    do {                                                                       \
        char *err_str = strerror_r((errno), (buf), (len));                     \
        /* If return value _isn't_ the start of the buffer we passed in,       \
         * then GNU strerror_r returned an internal static buffer and we       \
         * need to copy the result into our private buffer. */                 \
        if (err_str != (buf)) {                                                \
            buf[(len)-1] = '\0';                                               \
            strncat((buf), err_str, ((len) - 1));                              \
        }                                                                      \
    } while (0)
```

badboy:
strncat might copy n+1 bytes (n bytes from the source plus a terminating nul byte).
Also strncat appends after the first found nul byte. But all we pass is
a buffer we might not have zeroed out already.

Closes #380

### DIFF
--- a/hiredis.h
+++ b/hiredis.h
@@ -98,8 +98,8 @@
          * then GNU strerror_r returned an internal static buffer and we       \
          * need to copy the result into our private buffer. */                 \
         if (err_str != (buf)) {                                                \
-            buf[(len)] = '\0';                                                 \
-            strncat((buf), err_str, ((len) - 1));                              \
+            strncpy((buf), err_str, ((len) - 1));                              \
+            buf[(len)-1] = '\0';                                               \
         }                                                                      \
     } while (0)
 #endif


badboy:
@not-a-robot r+

PR #404, opened 2016-03-21
Update README.md to tell about SUBSCRIBE limitations with redisAsyncContext

Gui13:
When issuing a SUBSCRIBE command to a `redisAsyncContext`, you cannot send any other command than (P)SUBSCRIBE/UNSUBSCRIBE, until the context has no pending subscriptions.

This updates the README to tell people using HiRedis about this limitation, and the way to compensate for it (by using 2 contexts). The API doesn't hint at this even though it looks like a legitimate usage (place a subscription, expect to be called when a publish is issued, and use the same context to do other operations).

Feel free to enhance my prose..

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,6 +299,23 @@ valid for the duration of the callback.
 
 All pending callbacks are called with a `NULL` reply when the context encountered an error.
 
+#### A note on SUBSCRIBE/UNSUBSCRIBE
+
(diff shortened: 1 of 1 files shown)


badboy:
That's actual a limitation of Redis.

badboy:
I'm not going to pull this in. It's a limitation of Redis, not hiredis itself. Our documentation is already quite huge and unstructured. 

If we ever get a better structured documentation, we might have a section on potential problems like these.

PR #406, opened 2016-04-11
Update read.c

DongwenHuang:
static char *seekNewline(char *s, size_t len)  : 
this function can not parse the string,such as "hello world\r". the case that  the last char is '\r'.


badboy:
@not-a-robot

badboy:
@not-a-robot r

badboy:
@not-a-robot r+

PR #412, opened 2016-04-20
Add C++ example and move old-style cast inlined function into sds.c

### LINKED ISSUE #364: static inline functions are not within extern C
Hi all,

Actually the complete sds.h is not inside an `extern C {`. This causes issues when hiredis is included in C++.

Why not put the extern C at the top of hiredis.h?

https://github.com/redis/hiredis/blob/master/hiredis.h#L107-L109

Thanks,

Rik

xor-gate:
This fixes #364 
Probably static inlining is not a good idea because the compiler is much smarter: 

https://www.kernel.org/doc/Documentation/CodingStyle
Chapter 15: The inline disease

I think because sds actually comes from redis itself it needs to be fixed there and be pulled into hiredis.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 # This file is released under the BSD license, see the COPYING file
 
 OBJ=net.o hiredis.o sds.o async.o read.o
-EXAMPLES=hiredis-example hiredis-example-libevent hiredis-example-libev hiredis-example-glib
+EXAMPLES=hiredis-example hiredis-example-cpp hiredis-example-libevent hiredis-example-libev hiredis-example-glib
 TESTS=hiredis-test
 LIBNAME=libhiredis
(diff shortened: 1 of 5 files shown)


badboy:
I'm trying to sync up with the upstream sds library (and Redis) and thus can't pull it in as is. sds still defines sdslen and sdsavail as `static inline`, thus we will keep it like this here as well.

rikvdh:
@badboy How do you want to integrate the new SDS?
Do you want it as a submodule? Or copy-paste and hold duplicates everywhere? Perhaps I can help integrating the new SDS, because I get the feeling that changes to hiredis are kept back because "new SDS is coming"..

xor-gate:
+1 for better traceable git submodule

PR #422, opened 2016-05-09
Changes for IBM AIX systems.

kucik:
As I tried to use redis api library on AIX system, there are some minor things I had to change.
1. AF_UNIX should be supported on most systems instead of AF_LOCAL
2. _XOPEN_SOURCE_EXTENDED 1 to enable struct timeval in sys/time.h
3. _ALL_SOURCE for netdb.h to enable struct addrinfo.
4. _XOPEN_SOURCE because of:
   
   struct timeval {
       time_t      tv_sec;         /\* seconds */
   # if defined(**64BIT**) || (_XOPEN_SOURCE >= 500)
   
     suseconds_t tv_usec;        /\* microseconds */
   # else
   
   long        tv_usec;        /\* microseconds */
   # endif
   
   };

### DIFF
--- a/fmacros.h
+++ b/fmacros.h
@@ -15,11 +15,19 @@
 #elif defined(__linux__) || defined(__OpenBSD__) || defined(__NetBSD__)
 #define _XOPEN_SOURCE 600
 #else
+#ifndef AIX
 #define _XOPEN_SOURCE
+#endif
 #endif
 
 #if defined(__APPLE__) && defined(__MACH__)
 #define _OSX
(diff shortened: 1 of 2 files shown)


badboy:
Triggering CI.

badboy:
For Sun systems we already define `AF_LOCAL` to be `AF_UNIX`: https://github.com/redis/hiredis/blob/master/net.h#L41

We can probably just change it to `AF_UNIX`, that seems to be more portable anyway.

PR #426, opened 2016-05-13
Typo format in redisFormatSdsCommandArgv function

paulinohuerta:
this issue is very significant, because not allow the proper execution of the "function redisCommandArgv". The server returns "invalid bulk length".
Thanks!


badboy:
Thanks, I added a test in #426

PR #428, opened 2016-05-14
Update sds.h

avatli:
Fixing sds.h for building hiredis in cpp project


xor-gate:
This will not be merged as seen the same issue for PR #412

PR #429, opened 2016-05-20
using new version libevent

xinchuantao:


badboy:
Why the type changes?

badboy:
Ok, checked the code. This is good!

@not-a-robot r+

PR #430, opened 2016-05-23
Fix a compile error for C++

maxilai:
There is a compile error in the file sds.h when using g++. 
error: invalid conversion from ‘void*’ to ...

This issue is fixed by this commit.

Thanks!


maxilai:
The same issue has been raised here:  Update sds.h #428

PR #433, opened 2016-05-30
net: Fixes #431 compile issue on osx with AF_LOCAL and TCP_KEEPALIVE

### LINKED ISSUE #431: osx make failed
```
cc -std=c99 -pedantic -c -O3 -fPIC  -Wall -W -Wstrict-prototypes -Wwrite-strings -g -ggdb  net.c
net.c:139:37: error: use of undeclared identifier 'TCP_KEEPALIVE'
    if (setsockopt(fd, IPPROTO_TCP, TCP_KEEPALIVE, &val, sizeof(val)) < 0) {
                                    ^
net.c:433:29: error: use of undeclared identifier 'AF_LOCAL'
    if (redisCreateSocket(c,AF_LOCAL) < 0)
                            ^
net.c:458:21: error: use of undeclared identifier 'AF_LOCAL'
    sa.sun_family = AF_LOCAL;
                    ^
3 errors generated.
make: *** [net.o] Error 1
```

/usr/include/netinet  exists

clang -v

```
Apple LLVM version 7.3.0 (clang-703.0.31)
Target: x86_64-apple-darwin15.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

xor-gate:
It also enable building on OS X on Travis CI so future builds will be catched early after push.

This is related to: 
https://github.com/joyent/libuv/issues/228
https://github.com/nviennot/nginx-tcp-keepalive/issues/2

Fixes #431

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,19 @@ addons:
     - gcc-multilib
     - valgrind
 
+before_script:
+    - if [ "$TRAVIS_OS_NAME" == "osx" ] ; then brew update; brew install redis; fi
+
 env:
     - CFLAGS="-Werror"
     - PRE="valgrind --track-origins=yes --leak-check=full"
     - TARGET="32bit" TARGET_VARS="32bit-vars" CFLAGS="-Werror"
     - TARGET="32bit" TARGET_VARS="32bit-vars" PRE="valgrind --track-origins=yes --leak-check=full"
+
+matrix:
+  include:
+    - os: osx
+      compiler: clang
+      env: CFLAGS="-Werror" PRE=""
 
 script: make $TARGET CFLAGS="$CFLAGS" && make check PRE="$PRE" && make $TARGET_VARS hiredis-example
--- a/net.c
+++ b/net.c
@@ -135,12 +135,6 @@ int redisKeepAlive(redisContext *c, int interval) {
 
     val = interval;
 
-#ifdef _OSX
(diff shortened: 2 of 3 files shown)


badboy:
Interesting. Did this change recently? I'm pretty sure people ran it on OS X before, though I don't use it so can't tell.

xor-gate:
As it doesn't compile, it must have been changed in some way. I'm not using it either but this is not handy it breaks silently. Travis has support for OS X building for people who don't own macintosh machines -_-".

badboy:
Yeah, definitely. Having CI run on OSX can't hurt. I just want to understand why it broke.

xor-gate:
Yeah I understand, we need to dig a little deeper **why** it is broken currently.

xor-gate:
Issue #431 is closed but probably this PR fixes also the issue for others, seperate from the problem when it was introduced. We should look forward now and make it work for users, and look backward in git commit history when it works again.

tnm:
@badboy:
> Interesting. Did this change recently? I'm pretty sure people ran it on OS X before, though I don't use it so can't tell.

From a quick bisect, this appears to be the breaking change: https://github.com/redis/hiredis/pull/378/commits/bb1747b1bf92431e5c5e9699824d2ef52f863f45 in https://github.com/redis/hiredis/pull/378.

badboy:
Thanks, reverting that change makes more sense I guess.
Will test it together with the Travis changes later today and hopefully pull it in.

badboy:
Closing in favor of #439

PR #435, opened 2016-05-30
Fixing compiler warnings

rgowrishankar:
With some flags enabled, the warnings that all declarations should precede code gets triggered. Fixing it by moving declarations to top of function.


badboy:
Not a fan of moving declarations to the top. Keep it where it is used.

PR #445, opened 2016-07-07
fix: should close socket fd when retry connect (tcp)

xxvv:


badboy:
I guess it makes sense. I think there is no proper way to test this reliable though.

badboy:
```
APPLICATION USAGE
       If connect() fails, the state of the socket is unspecified. Conforming applications should close the file descriptor and create a new socket before attempting to reconnect.
```

Perfect.

@not-a-robot r+

PR #451, opened 2016-07-25
Include update to handle latest Nidium server

fredericandre:


badboy:
Please include an example and a target in the Makefile to compile it.
I have never heard nor used libapenetwork (what is nidium?) and so have no way to ensure that the adapter keeps working.

PR #455, opened 2016-08-14
NULL pointer dereferences fixed

tpltnt:
Hi there,

I fixed some potential NULL pointer dereferences.
Cheers,
tpltnt


badboy:
We are currently missing _a lot_ of null checks, primarily because when used in Redis it is no problem (as malloc is aliased and handles the NULL return anyway).

Plus we don't do null checks in the test at all, so fixing just one there makes even less sense.

sinisterstuf:
It seems like although it is with good intent, what this pull request aims to fix is not actually a problem, should we then close the pull request?

PR #461, opened 2016-09-01
Libuv polling error reporting bugfix

avax94:
Hey everyone,
So redisAsyncConnect wasn't giving any information when trying to connect to bad host, or when connection is closed using libuv as an adapter.

This should fix it.

Thank you


michael-grunder:
Closing as #578 took care of it

PR #468, opened 2016-09-19
Prevented uv adapter from calling write when context has been freed

paul-scott:
The `redisLibuvPoll` function can be called with both the `UV_READABLE` and `UV_WRITABLE` flags set at the same time.  Calling `redisAsyncHandleRead` can lead to a disconnect and the context being cleaned up/freed.  If this happens then `redisAsyncHandleWrite` should not be called otherwise memory read/write errors and duplicate freeing will occur.

These changes prevent this from happening by having the `redisLibuvCleanup` callback indicate that the context has been cleaned.  This is done indirectly by setting the context to a null pointer, maybe someone can come up with a cleaner way.

### DIFF
--- a/adapters/libuv.h
+++ b/adapters/libuv.h
@@ -20,10 +20,10 @@ static void redisLibuvPoll(uv_poll_t* handle, int status, int events) {
     return;
   }
 
-  if (events & UV_READABLE) {
+  if (p->context != NULL && (events & UV_READABLE)) {
     redisAsyncHandleRead(p->context);
   }
-  if (events & UV_WRITABLE) {
+  if (p->context != NULL && (events & UV_WRITABLE)) {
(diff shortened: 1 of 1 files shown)


badboy:
This kinda prevents reconnecting, but we don't have in-built reconnecting for async contexts anyway, so we can make that change for now.

badboy:
@not-a-robot r+

kelvinh:
As the newest release 0.13.3 does not contain this fix, is there a plan to make a new release to contain this fix? We are also issuing the double free problem during disconnecting, most time it's OK, only occasionally the program crashes with SIGSEV, so it's hard to debug, due to many developers always use release version but not the master branch, I think make a new release will save a lot of debugging time for other developers.

Issue #396, opened 2016-02-17
async.c assert on rapid re-subscription


andreca:
The assert (https://github.com/redis/hiredis/blob/master/async.c#L463) is triggered when a PSUBSCRIBE is called "right after" a PUNSUBSCRIBE (to the same pattern).

If I postpone the second PSUBSCRIBE with 1 second (`i == 4`) the assert is not triggered.

Example code:

``` c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <hiredis/async.h>
#include <hiredis/adapters/libevent.h>

int main (int argc, char **argv)
{
  struct event_base *base = event_base_new();
  redisAsyncContext *c = redisAsyncConnect("127.0.0.1", 6379);
  if(c == NULL || c->err)
  {
    printf("error creating context\n");
    return 1;
  }

  redisLibeventAttach(c, base);

  for(size_t i = 0; i < 5; ++i)
  {
    if(i == 0)
    {
      if(redisAsyncCommand(c, NULL, NULL, "PSUBSCRIBE test") != REDIS_OK)
        printf("error: %s\n", c->errstr);
    }
    else if(i == 2)
    {
      if(redisAsyncCommand(c, NULL, NULL, "PUNSUBSCRIBE test") != REDIS_OK)
        printf("error: %s\n", c->errstr);
    }
    else if(i == 3)
    {
      if(redisAsyncCommand(c, NULL, NULL, "PSUBSCRIBE test") != REDIS_OK)
        printf("error: %s\n", c->errstr);
    }

    sleep(1);
    event_base_loop(base, EVLOOP_ONCE);
  }
  return 0;
}
```

the output is:

```
test: async.c:463: redisProcessCallbacks: Assertion `(c->flags & 0x20 || c->flags & 0x40)' failed.
```

Viish:
I'm getting the same behavior when doing multiple UNSUBSCRIBE in a short period of time:
> flexisip: async.c:411: redisProcessCallbacks: Assertion `(c->flags & 0x20 || c->flags & 0x40)' failed.

NKovner:
We see the same thing and confirmed it happens in the case described by Viish.

PR #472, opened 2016-09-30
Counting pending subscribe to unset SUBSCRIBED flag properly in async context.

hyjin:
Issue #396 shows assertion fail on re-subscribing right after unsubscribe command.

It is caused after we unset `REDIS_SUBSCRIBED` flag on unsubscribe reply with 0 subscription, but we still have pending subscribe reply, and assertion fails the context's `REDIS_SUBSCRIBED` flag if off.

To evade this, I introduced a pending subscribe counter to the `struc tredisAsyncCallback` in order to unset the flag in a right time.

### DIFF
--- a/async.c
+++ b/async.c
@@ -395,7 +395,7 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
                 cb->pending_subs -= 1;
             }
 
-            memcpy(dstcb,dictGetEntryVal(de),sizeof(*dstcb));
+            memcpy(dstcb,cb,sizeof(*dstcb));
 
             /* If this is an unsubscribe message, remove it. */
             if (strcasecmp(stype+pvariant,"unsubscribe") == 0) {
--- a/async.c
+++ b/async.c
@@ -364,6 +364,7 @@ void redisAsyncDisconnect(redisAsyncContext *ac) {
 static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply, redisCallback *dstcb) {
     redisContext *c = &(ac->c);
     dict *callbacks;
+    redisCallback *cb;
     dictEntry *de;
     int pvariant;
     char *stype;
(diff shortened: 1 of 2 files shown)


async.c
```diff
@@ -387,16 +388,26 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
...
+                if (cb->pending_subs == 0)
+                    dictDelete(callbacks,sname);
 
                 /* If this was the last unsubscribe message, revert to
                  * non-subscribe mode. */
                 assert(reply->element[2]->type == REDIS_REPLY_INTEGER);
-                if (reply->element[2]->integer == 0)
+
+                /* Unset subscribed flag only when no pipelined pending subscribe. */
+                if (reply->element[2]->integer == 0 && dictSize(callbacks) == 0)
```
tvroegop: The redisAsyncContext can have pattern as well as channel subscriptions. The REDIS_SUBSCRIBED flag should be unset only if both are all gone.

async.c
```diff
@@ -387,16 +388,26 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
...
         de = dictFind(callbacks,sname);
         if (de != NULL) {
+            cb = dictGetEntryVal(de);
+
+            /* If this is an subscribe reply decrease pending counter. */
+            if (strcasecmp(stype+pvariant,"subscribe") == 0) {
+                cb->pending_subs -= 1;
+            }
+
             memcpy(dstcb,dictGetEntryVal(de),sizeof(*dstcb));
```
tvroegop: dictGetEntryVal(de) -> cb

async.c
hyjin: That's right. I missed that.

async.c
hyjin: You are correct. I should have change this line to use the `cb`, too.

hyjin:
@tvroegop Thank you for the review. I've fixed where you pointed out.

winnieelte:
Hi guys! Do you have any estimation when will this change go back to master?

michael-grunder:
Hey,

I apologize for the delayed reply but I completely missed this PR when it was opened.  I'll build it locally and do some testing tomorrow but at first glance, it looks good.

winnieelte:
Hi, any update about this? Thanks in advance!

michael-grunder:
Merged, thanks so much!

PR #473, opened 2016-10-09
Build error when make examples , patch with glib-2.0 positon move will solve.

trustfarm-dev:
In case of some glib-2.0 linker error ,
make examples 
can't link with glib2.0,  in this case -lglib-2.0  to after includes and move to last will solve the issues.


badboy:
This just re-orders the arguments. When did you encounter a problem with the current order?

PR #479, opened 2016-10-25
Remove trailing comma in redisConnectionType enum

DonaldWhyte:
Remove trailing comma in last value of `redisConnectionType` enum. This causes a compiler warning on Solaris compilers. I'd like to build this on Solaris with `-Werror`. However, due to the trailing comma, I cannot do that.

This PR removes the trailing comma, which should prevent it causing compiler warnings on any architecture.

### DIFF
--- a/hiredis.h
+++ b/hiredis.h
@@ -133,7 +133,7 @@ void redisFreeSdsCommand(sds cmd);
 
 enum redisConnectionType {
     REDIS_CONN_TCP,
-    REDIS_CONN_UNIX,
+    REDIS_CONN_UNIX
 };
 
 /* Context for a connection to Redis */


badboy:
Who the hell made trailing commas in enum definitions a warning on Solaris?

DonaldWhyte:
Yes, it's an interesting choice. This offers an explanation:

http://stackoverflow.com/questions/6372650/trailing-commas-and-c

The Solaris compiler used is C++03 (a fully C++11-conformant Solaris compiler currently does not exist sadly). C++03 tries to be backwards compatible with C89, with does *not* allow trailing commas.

C99 does allow trailing commas, and only C++11 has caught up with C99 by also allowing them.

As long as we're stuck with C++03-ish compilers for Solaris, looks like trailing commas will be an issue.

DonaldWhyte:
Is this good to land?

badboy:
I guess, yes.

DonaldWhyte:
Thanks @badboy!

PR #489, opened 2016-11-25
External memory management

MarkRohrbacher:
These patches allow to use a custom memory management without overwriting malloc / free et al.


badboy:
I'm not gonna pull this in, as I don't see the benefit of having an external memory management as part of this library.

PR #504, opened 2017-01-17
Watch out for memory leak in cases where allocation fails and leaks the original memory that's no longer referred to

aaronjwood:
Originally submitted here https://github.com/antirez/redis/pull/3439

### DIFF
--- a/sds.c
+++ b/sds.c
@@ -1018,7 +1018,14 @@ sds *sdssplitargs(const char *line, int *argc) {
                 if (*p) p++;
             }
             /* add the token to the vector */
-            vector = s_realloc(vector,((*argc)+1)*sizeof(char*));
+            char **data = s_realloc(vector,((*argc)+1)*sizeof(char*));
+            if (data == NULL) {
+                /* Return something not NULL. */
+                return s_malloc(sizeof(void*));
+            }
+            else {
+                vector = data;
+            }
             vector[*argc] = current;
             (*argc)++;
             current = NULL;


sds.c
```diff
@@ -1018,7 +1018,14 @@ sds *sdssplitargs(const char *line, int *argc) {
                 if (*p) p++;
             }
             /* add the token to the vector */
-            vector = s_realloc(vector,((*argc)+1)*sizeof(char*));
+            char **data = s_realloc(vector,((*argc)+1)*sizeof(char*));
+            if (data == NULL) {
+                /* Return something not NULL. */
+                return s_malloc(sizeof(void*));
```
badboy: Why return somethign non-NULL here? Wouldn't this break users of this code as they have to assume the function worked just fine?

aaronjwood:
I had initially done this because I found other parts of redis around where my change was did the same thing. Do you have a macro or something similar to handle these situations that immediately exits?

sds.c
badboy: What part of the code would re-try an allocation in order to return this memory uninitialized (if it succeeds)?
That looks like a serious bug waiting to happen from the user-perspective. How would the user even know that this memory does not hold any valid data?

aaronjwood: I agree with what you're saying. I had put this as part of my patch because hiredis does it https://github.com/antirez/redis/blob/unstable/deps/hiredis/sds.c#L1027

I can certainly change this. I just don't have enough context to assume what the maintainers would want.

PR #514, opened 2017-03-28
redisLibeventCleanup will be leak

zfzsoft:
event_del can not free the "e->rev" and "e->wev",that will leak when reconnect the redis


badboy:
`rev` and `wev` are embeded into our `redisLibeventEvents` and thus not individually allocated. They can't be freed that way. They will be freed just one line further down when freeing the event struct.

badboy:
Woops, I looked at the old code. You are right, we now carry only pointers and allocate them. Merging this here.

PR #515, opened 2017-03-30
fix memory leak: free libevent events on cleanup

jasonish:
The event's were being removed from the event loop but never freed.


jasonish:
Woops. Someone already did this.. https://github.com/redis/hiredis/pull/514

PR #524, opened 2017-05-15
Don't pass a negative value to __redisAsyncCommand if redisFormatSdsCommandArgv fails

deweerdt:


michael-grunder:
Looks good to me, nice find!

deweerdt:
I'm confused by https://github.com/redis/hiredis/pull/524#issuecomment-315538991 it does look like the code landed on master?

badboy:
That's what you get for mixing mutliple bots into one repo. The change landed, the @not-a-robot user is legacy.

deweerdt:
:+1: thanks for the explanation.

PR #532, opened 2017-06-15
Assert statement calls a function which has side effects

amallia:


badboy:
Oh, indeed. Waiting for tests to finish, then merging

amallia:
@badboy should be fine to merge

PR #533, opened 2017-06-15
Small fixes

amallia:


badboy:
Gnah, I realized I never send off these comments.

hiredis.c
```diff
@@ -1007,9 +1007,8 @@ void *redisvCommand(redisContext *c, const char *format, va_list ap) {
 
 void *redisCommand(redisContext *c, const char *format, ...) {
     va_list ap;
-    void *reply = NULL;
     va_start(ap,format);
-    reply = redisvCommand(c,format,ap);
+    void *reply = redisvCommand(c,format,ap);
```
amallia: Why not?

net.c
```diff
@@ -143,7 +143,6 @@ int redisKeepAlive(redisContext *c, int interval) {
     }
 #else
 #if defined(__GLIBC__) && !defined(__FreeBSD_kernel__)
-    val = interval;
```
amallia: Because it is already set before
```cpp
     val = interval;
 		 
 #ifdef _OSX		 #ifdef _OSX
     if (setsockopt(fd, IPPROTO_TCP, TCP_KEEPALIVE, &val, sizeof(val)) < 0) {		     if (setsockopt(fd, IPPROTO_TCP, TCP_KEEPALIVE, &val, sizeof(val)) < 0) {
         __redisSetError(c,REDIS_ERR_OTHER,strerror(errno));		         __redisSetError(c,REDIS_ERR_OTHER,strerror(errno));
         return REDIS_ERR;		         return REDIS_ERR;
      }		      }
  #else		  #else
  #if defined(__GLIBC__) && !defined(__FreeBSD_kernel__)		  #if defined(__GLIBC__) && !defined(__FreeBSD_kernel__)
 -    val = interval;		
```

badboy:
I don't mind too much.

bors r+

PR #536, opened 2017-06-30
redisSetBlocking should not always be unblocking when REDIS_BLOCK is set.

binnnliu:


badboy:
We reset it a few lines further down again: https://github.com/redis/hiredis/blob/4d8f3fb207aa63a44999f78bc089a19c846eb9a9/net.c#L396
We handle the blocking case just above that (`redisContextWaitReady` in case `connect` returned `EINPROGRESS`).
With a blocking socket we can't guarantee the timeout.

PR #539, opened 2017-08-16
calloc param fixes and NULL check

cdliuqiang:
1、The calloc function prototype is void *calloc(size_t nmemb, size_t size)， 
Correct call is calloc(1,sizeof(redisReader));

2、redisContext *c = redisContextInit() maybe return NULL,  check it;


badboy:
Merged as 042b99b269f66a049aaf1eebe73d7ef71b151a60

PR #554, opened 2017-10-30
build: do not assume that INSTALL is cp

ignatenkobrain:
INSTALL is supposed to be `install` in most of the cases which
doesn't work with directories, but works perfectly with files.

Don't do this assumption.

Reported-by: Jiří Vymazal <jvymazal@redhat.com>
References: https://bugzilla.redhat.com/show_bug.cgi?id=1506251
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


badboy:
bors r=

PR #558, opened 2017-11-06
Prevent FD leakage when calling fork(2)+exec(2) 

### LINKED ISSUE #555: File descriptor handlers leak on fork + exec
When `fork`ing and `exec`ing, file descriptors get leaked to the child process, which is not always ideal. One can always call `fnctl` and set `O_CLOEXEC` after establishing a connection, but this can lead to race conditions if a different thread is trying to `fork`.

Apart from OS compatibility, are there any cases where one might want this sort of behaviour, meaning the sockets couldn't be created with `SOCK_CLOEXEC`? If so, how could we implement some sort of flag or another `Connect*` in order to accommodate for that?

I've been bitten by this at work and I went with calling `fnctl`, but I'd like to have a more atomic fix, and I'm willing to contribute a patch to this, I'd just like to have some input on what sort of API/behaviour is best. :)

feliperuiz:
I couldn't think of a better way to test this behavior, but I'm open to suggestions.

This should resolve #555.

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -724,6 +724,47 @@ redisContext *redisConnectBindNonBlockWithReuse(const char *ip, int port,
     return c;
 }
 
+redisContext *redisConnectCloseOnExec(const char *ip, int port) {
+    redisContext *c;
+
+    c = redisContextInit();
+    if (c == NULL) {
+        return NULL;
+    }
+
+    c->flags |= REDIS_BLOCK;
+    c->flags |= REDIS_CLOEXEC;
+    redisContextConnectTcp(c,ip,port,NULL);
+    return c;
+}
+
+redisContext *redisConnectWithTimeoutCloseOnExec(const char *ip, int port, const struct timeval tv) {
+    redisContext *c;
+
+    c = redisContextInit();
+    if (c == NULL)
+        return NULL;
+
+    c->flags |= REDIS_BLOCK;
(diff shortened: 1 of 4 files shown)


stuz5000:
Does this fix resolve a leaking file descriptor issue, or just test for it?

I'm tracking down an issue with redis (every so often, my client can't connect to redis for long periods). 
I was worried that redis was leaking connections, perhaps this is the case. I've shut down all the redis client process on a client host, yet when I run lsof, I see many open files (>15) to the client host.

```
    "COMMAND": "redis-ser", 
    "DEVICE": "143909", 
    "FD": "43u", 
    "NAME": "redis.mycompany.com:6379->client.mycompany.com:34660", 
    "NODE": "TCP", 
    "PID": "869", 
    "SIZE/OFF": "0t0", 
    "STATUS": "(ESTABLISHED)", 
    "TYPE": "IPv4", 
    "USER": "redis"
```

In total ~100 open files (including connections to other machines. Not sure if this the cause of my socket problem -- although it could be if redis also enforces a #sockets cap per client (and fails to release dead connections).

zuiderkwast (commented):
This is good stuff.

I think it should be rebased/rewritten using redisConnectWithOptions.

michael-grunder:
I'll merge the commits and rework the logic to use `redisConnectWithOptions`

PR #564, opened 2017-12-27
Remove $(ARCH) expansion from the Makefile

dmvo:
Some compilation environments (such as Yocto) define the ARCH
environment variable to indicate the target architecture. For
such enviroments, hiredis build fails, because the expanded
$(ARCH) variable in the Makefile gets erroneously interpreted
as an argument to the `-ggdb' command line option during the
compilation stage or as an input file name during the linking
stage.

This patch removes $(ARCH) expansions from the Makefile. This
doesn't harm cross-compilation, the latter goes fine with the
properly assigned CC environment variable. For native builds,
this patch does not imply any changes.

Signed-off-by: Dmitri Vorobiev <dmitri.vorobiev@movial.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,8 +36,8 @@ endef
 export REDIS_TEST_CONFIG
 
 # Fallback to gcc when $CC is not in $PATH.
-CC:=$(shell sh -c 'type $(CC) >/dev/null 2>/dev/null && echo $(CC) || echo gcc')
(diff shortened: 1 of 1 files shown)


badboy:
Thanks! I traced this back to when Redis actually passed `ARCH=...` to force a 32-bit build. This is long changed.

PR #570, opened 2018-01-09
SSL Client support

mnunberg:
This provides SSL client support for communicating with an SSL-secured Redis implementation. The user-facing API is exposed as bolted-on, which makes modifying existing applications easier. This works in my basic testing so far.

The SSL implementation assumes mutual TLS auth. SSL support is disabled by default at compilation. To use SSL, set `USE_SSL` in the build environment.

```
$make USE_SSL=1
```

This should enable the `HIREDIS_SSL` preprocessor define, which should enable the SSL bits. Note that you may want to also adjust your `OPENSSL_PREFIX` build variable to your desired openssl installation. This is probably only a concern for OSX where there are typically both Homebrew and Apple variants provided. On linux there should only be a single version found in `/usr/lib/` or similar. The default value is `/opt/local/openssl`.

Once the library is built, you should be able to call

```
int redisSecureConnection(redisContext *c, const char *capath, const char *certpath,
                          const char *keypath);
```

On the connection. This will perform openssl negotiation. The reason I didn't provide another variant of `redisConnect` is twofold. (1) I wanted other applications to be able to simply add (rather than change) their existing connection code - so this also works via fd etc. (2) Users can quickly debug if an error is coming from the connection layer or openssl layer.

I've also added examples (the sync example and the libev and libevent examples).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,9 @@ STLIBNAME=$(LIBNAME).$(STLIBSUFFIX)
 STLIB_MAKE_CMD=ar rcs $(STLIBNAME)
 
 ifdef USE_SSL
-	OPENSSL_PREFIX=/usr/local/opt/openssl
+	# This is the prefix of openssl on my system. This should be the sane default
+	# based on the platform
+	OPENSSL_PREFIX?=/usr/local/opt/openssl
 	CFLAGS+=-I$(OPENSSL_PREFIX)/include -DHIREDIS_SSL
 	LDFLAGS+=-L$(OPENSSL_PREFIX)/lib -lssl -lcrypto
 endif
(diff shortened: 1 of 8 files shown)


mnunberg:
@michael-grunder Any thoughts on this? I've been asked to revive this

michael-grunder:
I'm not opposed to adding SSL support.

There was a great deal of back and forth around this issue in Redis proper, so we may want to run it by @antirez to see if he has any objections.

I'm happy to help test as well.

aa7955:
Is there any future for this?

mnunberg:
Yes, in fact SSL is actively being used in some deployments! - however I need to cut out some time to properly integrate it in hiredis for public consumption.

mnunberg:
I'm closing this PR in favor of https://github.com/redis/hiredis/pull/645

PR #578, opened 2018-03-04
Proper error reporting for connect failures

mnunberg:


mnunberg:
Note that another PR (https://github.com/redis/hiredis/pull/461) also exists. Unlike the linked PR, this PR doesn't add a new API, and it doesn't break the abstraction layer between socket and context.

mnunberg:
I'd also like to abstract getaddrinfo to be non-blocking; right now it's blocking and in async mode we never really rotate. I'd also like to abstract the sockaddr member as well.

mnunberg:
@dvirsky

dvirsky:
Cc: @yossigo can you review this?

net.c
```diff
@@ -375,6 +399,15 @@ static int _redisContextConnectTcp(redisContext *c, const char *addr, int port,
                 goto error;
             }
         }
+
+        /* For repeat connection */
+        if (c->saddr) {
+            free(c->saddr);
+        }
+        c->saddr = malloc(sizeof(*p->ai_addr));
```
yossigo: The allocation should be done based on `ai_addrlen`, as `ai_addr` is a vanilla sockaddr which may be smaller than what you really need (e.g. for AF_UNIX or AF_INET6).

mnunberg: oops - done. There's also `struct sockaddr_storage`, but ideally I'd like to preserve the original `addrinfo` structure anyway.

mnunberg:
Still working on getting the tests to pass..

PR #580, opened 2018-03-17
fix common realloc mistake and add null check more

charsyam:
1] hiredis is client library, so need more null check and memory lick check.
 because using null ptr can destroy application.

2] I refered https://www.mantidproject.org/How_to_fix_CppCheck_Errors for realloc issue.


mnunberg:
Any reason we can't merge this? Seems fine to me. Though, fighting against NULL returns from malloc is always an uphill battle..

michael-grunder:
Looks OK to me.  Like you say once you start getting NULL from malloc you probably have bigger problems, but it's certainly more correct.

PR #592, opened 2018-05-01
Fix some POSIX usage issues

justinbrewer:
Here's a set of patches fixing some POSIX usage issues. Looking at the history, there's been a lot of issues compiling on other platforms. The root of those problems is AF_LOCAL, the old name for the standardized AF_UNIX. Every platform seems to have their own nuances on whether to define AF_LOCAL. Replacing it with AF_UNIX, and stripping fmacros.h down to just set the two basic version macros fixes all but Darwin, which needs one extra just for TCP_KEEPALIVE.

In glibc, if _XOPEN_SOURCE or _POSIX_C_SOURCE are defined, features.h will not define _GNU_SOURCE, so only the standard version of strerror_r is exposed. __redis_strerror_r is dropped in favor of using the standard version directly. A check could be added to fmacros.h to produce an error if _GNU_SOURCE is defined, but the build will fail in either case.

Some redundancies involving calloc and free are also fixed. calloc is guaranteed to provide a zero-initialized buffer, so initializing fields to zero is redundant. free and various *free functions in hiredis are NULL-safe, so checking before calling them is redundant.

POSIX does not specify the behavior of freeaddrinfo(NULL). In glibc, it's a NOP, but on the BSDs it crashes, so a NULL-check is necessary. d4b715f0aa97bae7bf2f6ebfe4837f39c0451e84 added a code path that can call freeaddrinfo(NULL). The necessary check is added.

In addition to the automatic CI, this was also tested on FreeBSD, OpenBSD, and NetBSD. I don't have AIX or Solaris to test on, so I can't check those.

### DIFF
--- a/fmacros.h
+++ b/fmacros.h
@@ -1,38 +1,12 @@
 #ifndef __HIREDIS_FMACRO_H
 #define __HIREDIS_FMACRO_H
 
-#if defined(__linux__)
-#define _BSD_SOURCE
-#define _DEFAULT_SOURCE
-#endif
-
-#if defined(__CYGWIN__)
-#include <sys/cdefs.h>
-#endif
-
-#if defined(__linux__) || defined(__OpenBSD__) || defined(__NetBSD__)
 #define _XOPEN_SOURCE 600
-#elif defined(__APPLE__) && defined(__MACH__)
-#define _XOPEN_SOURCE
-#elif defined(__FreeBSD__)
-// intentionally left blank, don't define _XOPEN_SOURCE
-#elif defined(AIX)
-// intentionally left blank, don't define _XOPEN_SOURCE
-#else
-#define _XOPEN_SOURCE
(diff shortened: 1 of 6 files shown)


vyuh:
I could not build redis on FreeBSD 11.1 today due to undefined `AF_LOCAL` error during build of hiredis.

michael-grunder:
@justinbrewer I've merged your PR into a feature branch [here](https://github.com/redis/hiredis/tree/posix-build-fixes).

I've got a friend who may be able to verify the build on a Solaris VM for us.  Either way I'll leave this branch up for a few days and then merge it into master.

Thanks again!

michael-grunder:
I went ahead and spun up FreeBSD, Solaris 10 and Solaris 11 VMs and this PR actually fixed a bunch of build issues.

Merged into master, Thanks!

PR #595, opened 2018-05-14
Use absolute install_name

### LINKED ISSUE #437: Improper handling of install name for shared library on OS X
I am working on a project that has hiredis as a submodule in our git repo. Since we want to keep the build process contained in a single (user-owned) directory tree, we `make install` with the `make INSTALL="cp -R" PREFIX=$(pwd) install`. I then find the following incorrect behavior when I run otool.

```
[thomas@Mithlond] hiredis $ otool -D lib/libhiredis.dylib
lib/libhiredis.dylib:
libhiredis.0.13.dylib
```

Note that since the install name is incorrectly set, linking against this copy of the library, even by absolute path, will produce a binary that can't find libhiredis.dylib without manually altering LD_LIBRARY_PATH.

ryandesign:
The `install_name` on macOS must be an absolute path (or must use `@executable_path`, `@loader_path` or `@rpath`, but that's less common).

Fixes #437.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ endif
 ifeq ($(uname_S),Darwin)
   DYLIBSUFFIX=dylib
   DYLIB_MINOR_NAME=$(LIBNAME).$(HIREDIS_SONAME).$(DYLIBSUFFIX)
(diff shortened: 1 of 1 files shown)


michael-grunder:
Merged, thanks!

ryandesign:
Sorry for missing the trailing slash! Thanks for fixing that in 9d3b4ec703ee3d2d03920bc2dbb939d9598ca25d.

PR #596, opened 2018-05-14
Use cp -pPR instead of cp -a

ryandesign:
`cp` does not support the `-a` flag on all operating systems, such as Mac OS X Leopard 10.5.x or earlier, resulting in:

```
cp: illegal option -- a
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-pvX] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-pvX] source_file ... target_directory
make: *** [install] Error 64
```

From `man cp` on macOS Sierra 10.12.6, `cp -a` is exactly the same as `cp -pPR`, so use that instead.

From `man cp` on macOS Sierra:

> Historic versions of the `cp` utility had a `-r` option. This implementation supports that option; however, its use is strongly discouraged, as it does not correctly copy special files, symbolic links, or fifo's.

So don't ever use that.


michael-grunder:
Thanks, merged!

PR #597, opened 2018-05-21
Update createArray to take size_t

justinbrewer:
This makes createArray consistent with createString, which also takes size_t. Bounds-check and unit tests are updated to allow up to `min(SIZE_MAX,LLONG_MAX)`. Changelog is updated to mention this API break.

Also included is a missed Changelog entry mentioning the integer parsing changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 User code should compare this to `size_t` values as well.
 If it was used to compare to other values, casting might be necessary or can be removed, if casting was applied before.
 
+* `redisReplyObjectFunctions.createArray` now takes `size_t` for its length parameter.
+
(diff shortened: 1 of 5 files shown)


michael-grunder:
Good catch on 300fc01.  I went ahead and cherry-picked that into master.

The breaking change is trickier but it would appear that the main downstream repos [hiredis-rb](https://github.com/redis/hiredis-rb/tree/master/vendor), [hiredis-py](https://github.com/redis/hiredis-py/tree/master/vendor), and [hiredis-node](https://github.com/redis/hiredis-node/tree/master/deps) all pin to specific versions.

I figure we leave this up for a few days and then go ahead and get it merged if there are no objections.

PR #598, opened 2018-05-31
Use -dynamiclib instead of -shared on macOS

ryandesign:
`-dynamiclib` is the correct [documented](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/CreatingDynamicLibraries.html) flag to use to create dynamic libraries on macOS. Newer toolchains recognize `-shared` as a synonym of `-dynamiclib` but older toolchains don't.


michael-grunder:
Merged, thanks.

PR #620, opened 2018-10-24
Fix redisBufferRead documentation

hacst:
Stumbled over a small documentation issue while browsing the code. The redisBufferRead comment referred to redisContextReadReply which I cannot find in this codebase nor the old redis-tools one. Presumably this meant to say redisGetReplyFromReader which is how redisBufferRead is used in hiredis.c. Could've also meant the interface function redisReaderGetReply.


michael-grunder:
Merged, thanks!

PR #631, opened 2018-11-29
chore(doc): Using https instead of http in docs

Bambooin:


Bambooin:
This pull request is relative to:
https://github.com/antirez/redis/pull/5617
https://github.com/antirez/redis/pull/5621

mnunberg:
not really necessary...

Issue #31, opened 2011-04-03
hiredis.h is not compatible with C89 and C++98


agladysh:
When included in file, compiled with `--std=c89 --Wall --pedantic --Werror`, `hiredis.h` produces this error:

<pre>
lib/hiredis/hiredis.h:92: error: ISO C90 does not support ‘long long’
lib/hiredis/hiredis.h:111: error: ISO C90 does not support ‘long long’
</pre>

Similar error appears when compiling with `--std=c++98` (other flags as above).

<pre>
lib/hiredis/hiredis.h:92: error: ISO C++ 1998 does not support ‘long long’
lib/hiredis/hiredis.h:111: error: ISO C++ 1998 does not support ‘long long’
</pre>

I did not try to compile whole hiredis (yet).

<pre>
$ gcc --version
gcc (Ubuntu/Linaro 4.4.4-14ubuntu5) 4.4.5
Copyright (C) 2010 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
</pre>

PR #638, opened 2018-12-31
Enhanced error-handling (OOM) and more (redisAsync* family)

oshamash:
Hiredis: Advanced async-handling

+ hiredis.c: Convert redisSetError to take format and va_args
+ dict.*   : Fixed generic hash function -> cannot really take signed number
+ async.*  : Valdup function now handles OOM properly
             KeyCompare function uses proper sized integers
             nextArgument - change to use proper str-int conversion function
             Added redisAsyncAppend* function family to support buffer
                                                 optimizations in async-mode
             Added proper error handling on OOM errors [inc buffer rollback]

### DIFF
--- a/async.c
+++ b/async.c
@@ -29,7 +29,6 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-
 #include "fmacros.h"
 #include <stdlib.h>
 #include <string.h>
@@ -53,14 +52,14 @@ int __redisAppendCommand(redisContext *, const char *, size_t);
 
 /* Functions managing dictionary of callbacks for pub/sub. */
 static unsigned int callbackHash(const void *key) {
-    return dictGenHashFunction((const unsigned char *)key, sdslen((const sds)key));
+    return dictGenHashFunction(key, sdslen((const sds) key));
 }
 
 static void *callbackValDup(void *privdata, const void *src) {
     redisCallback *dup;
     ((void) privdata);
 
-    if (NULL != (dup = (redisCallback *)malloc(sizeof(*dup))))
+    if (NULL != (dup = malloc(sizeof(*dup))))
         memcpy(dup, src, sizeof(*dup));
 
     return dup;
@@ -100,7 +99,7 @@ static dictType callbackDict = {
 static redisAsyncContext *redisAsyncInitialize(redisContext *c) {
     redisAsyncContext *ac;
 
-    ac = (redisAsyncContext *)realloc(c,sizeof(redisAsyncContext));
+    ac = realloc(c,sizeof(redisAsyncContext));
     if (ac == NULL)
         return NULL;
 
@@ -144,6 +143,16 @@ static void __redisAsyncCopyError(redisAsyncContext *ac) {
     redisContext *c = &(ac->c);
     ac->err = c->err;
     ac->errstr = c->errstr;
+}
+
+static void __redisAsyncSetError(redisAsyncContext *ctx, int err_type, const char *msg) {
+    /* Set the errstr if not already set */
+    if (NULL == ctx->errstr) {
+        ctx->errstr = ctx->c.errstr;
+    }
+
+    ctx->err = err_type;
+    __redisSetError(&ctx->c, err_type, msg);
 }
 
 redisAsyncContext *redisAsyncConnectWithOptions(const redisOptions *options) {
@@ -328,7 +337,8 @@ void __redisAsyncDisconnect(redisAsyncContext *ac) {
 
     if (ac->err == 0) {
         /* For clean disconnects, there should be no pending callbacks. */
-        assert(__redisShiftCallback(&ac->replies,NULL) == REDIS_ERR);
+        int ret = __redisShiftCallback(&ac->replies,NULL);
+        assert(ret == REDIS_ERR);
     } else {
         /* Disconnection is caused by an error, make sure that pending
          * callbacks cannot call new commands. */
@@ -587,8 +597,6 @@ void redisAsyncHandleWrite(redisAsyncContext *ac) {
     c->funcs->async_write(ac);
 }
 
-void __redisSetError(redisContext *c, int type, const char *str);
-
 void redisAsyncHandleTimeout(redisAsyncContext *ac) {
     redisContext *c = &(ac->c);
     redisCallback cb;
@@ -599,7 +607,7 @@ void redisAsyncHandleTimeout(redisAsyncContext *ac) {
     }
 
     if (!c->err) {
-        __redisSetError(c, REDIS_ERR_TIMEOUT, "Timeout");
+        __redisAsyncSetError(ac, REDIS_ERR_TIMEOUT, "Timeout");
     }
 
     if (!(c->flags & REDIS_CONNECTED) && ac->onConnect) {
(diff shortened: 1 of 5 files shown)


mnunberg:
I'd prefer if these were separate pull requests or commits; I'm not entirel sure what all these changes are intended to do.

async.c
```diff
@@ -105,7 +111,7 @@ static dictType callbackDict = {
 static redisAsyncContext *redisAsyncInitialize(redisContext *c) {
     redisAsyncContext *ac;
 
-    ac = realloc(c,sizeof(redisAsyncContext));
+    ac = (redisAsyncContext *)realloc(c,sizeof(redisAsyncContext));
```
michael-grunder: Why typecast here?

oshamash: Just cause realloc returns a void pointer,
In general this is actually a partial commit of a greater one I'm doing as part of using hiredis for a small POC, first I would like to fix any bugs related to memory management since this POC will require large data quantities going from/to redis via hiredis [will try to reach 2/4GB buffering limits]

michael-grunder:
I haven't gone through everything yet, but this looks quite good.  There are a bunch of type fixes as well as better OOM handling.

Also, I'm guessing that the addition of `redisAsyncAppend` style commands is going to be a rather substantial performance improvement.

I was going to suggest adding tests for the new async append functionality, but we apparently don't have any async based tests :joy:

async.c
michael-grunder: The standard guarantees a void pointer can be converted into any other pointer type, so AFAIK there is no reason to typecast `malloc` class methods in C (since C89 maybe?).

C++ is another story :smile:

oshamash: I stand corrected, I guess it's just me being used to C++

mnunberg: Please remove the typecasts; it would make it much easier to read, too.

async.c
```diff
@@ -616,19 +621,40 @@ static int __redisAsyncCommand(redisAsyncContext *ac, redisCallbackFn *fn, void
...
     clen -= pvariant;
 
+    /* NOTICE:
+     *  Further non thread-safe actions [ we append to buffer before pushing having proper callback ]
+     *  Since hiredis is already not thread-safe let's just append first to check for OOM errors */
+
+    /* Make sure we can append command before making it's callback - Delegate error */
+    if (REDIS_OK != (ret = __redisAppendCommand(c,cmd,len))) return ret;
+
+#define SAFE_PUSHCALLBACK(cb_list) \
```
mnunberg: I'm not a fan of the "SAFE" prefix. Just do `PUSH_CALLBACK`.

async.c
```diff
@@ -653,25 +679,109 @@ static int __redisAsyncCommand(redisAsyncContext *ac, redisCallbackFn *fn, void
...
+#undef SAFE_PUSHCALLBACK
 
-    __redisAppendCommand(c,cmd,len);
+    return REDIS_OK;
 
-    /* Always schedule a write when the write buffer is non-empty */
-    _EL_ADD_WRITE(ac);
+oom_error:
+    /* Rollback the obuf [size still allocated] */
+    clen = sdslen(c->obuf) - len;
```
mnunberg: If reacting to an OOM, `sdsRemoveFreeSpace()` is helpful too. https://github.com/redis/hiredis/blob/master/sds.h#L257

async.c
```diff
@@ -681,10 +791,21 @@ int redisvAsyncCommand(redisAsyncContext *ac, redisCallbackFn *fn, void *privdat
     len = redisvFormatCommand(&cmd,format,ap);
 
     /* We don't want to pass -1 or -2 to future functions as a length. */
-    if (len < 0)
+    if (len == -1) {
```
mnunberg: Might want to `#define`-away the meaning of -1, -2, etc.

mnunberg: Might also want to do something like `redisvAsyncStrerror()` (static) rather than specify inline...

hiredis.c
```diff
@@ -563,15 +563,15 @@ void redisFreeCommand(char *cmd) {
...
     c->err = type;
-    if (str != NULL) {
-        len = strlen(str);
-        len = len < (sizeof(c->errstr)-1) ? len : (sizeof(c->errstr)-1);
-        memcpy(c->errstr,str,len);
-        c->errstr[len] = '\0';
+    if (format != NULL) {
+        va_start(args, format);
+        /* We ignore return value since SetError has no return code */
+        (void) vsnprintf(c->errstr, sizeof(c->errstr), format, args);
```
mnunberg: I am unsure if we are still aiming for C89 compliance. If we are, then `vsnprintf` is verbotten, since it is C99. Maybe @michael-grunder can tell us more.

mnunberg:
Looks good overall, but the numerous stylistic changes mixed in with edit changes makes it difficult to discern what's what in this commit. The biggest help, I think, would be a substitute function for all the boilerplate error code checking.

async.c
oshamash: Fixed

async.c
oshamash: Added call to sdsRemoveFreeSpace() in hopes it will help the users

async.c
oshamash: I would like that too but currently these are defined by the redisvFormatCommand() 
If we would like to use a define we should change it there and delegate properly,
currently I don't want to go over the entire code base and see if this value was delegated somewhere and make sure we change it project-wide, probably should be a different commit/PR

oshamash: > Might also want to do something like `redisvAsyncStrerror()` (static) rather than specify inline...

Added something similar to that

hiredis.c
oshamash: Reverted this change since I don't use it in this commit anyways

michael-grunder: It's no longer important vis-a-vis this commit but AFAIK hiredis has always required C99, so it's safe to do things that won't work in C89/C90.

mnunberg: Looking at the code, I see classic marks of C89, for example, no in-line declarations of variables, and no usages of designated initializers. If hiredis was ever intended to be compiled with VS 2008, then it's C89

michael-grunder: You could totally be right that hiredis once targeted C89, but if it did it was before I became aware of the project.

It seems to be C99 as far back as [4/4/2011](https://github.com/redis/hiredis/issues/31#issuecomment-952785)

I have no strong feelings on the matter but if we decided to support C89 we'd have to start maintaining two copies of things (e.g. sds strings).

mnunberg: good to know!

OmriSteiner:
I'm interested in the redisAsyncAppend* functionality in this branch.
@mnunberg If I strip this pull request of its stylistic changes and create a more concise pull request, would it be eligible for a merge?

oshamash:
@OmriSteiner @mnunberg Due to comment decided to rebase properly and remove needless typecasts, please review :)

michael-grunder:
The most recent changes look good to me, but I'm going to go through the entire diff of master because I only quickly scanned it before, plus it was quite some time ago.

I'm in favor of merging the changes but we'll want to coordinate with the Redis repo as it uses hiredis as a dependency and we should avoid causing pain there.

michael-grunder:
@oshamash We ended up handling OOM #800 but I merged your improvements involving the `redisAsyncAppend` logic [in a branch on my fork](https://github.com/michael-grunder/hiredis/tree/async-append).

I think the `redisAsyncAppend` commands will be nice to have.  Please let me know if I broke something in my merging the branch.

oshamash:
@michael-grunder nice, went over the last few commits there, deserves a big like, edge case handling like oom are highly in need when scaling up your redis game while keeping hiredis as your client of choice.
In general I might add sentinel support soon, but honestly I could not find a nice clean way to use it (pm for more info)

michael-grunder:
Closing this in favor of #822

Hopefully we can get it merged before v1.0.0

PR #641, opened 2019-01-13
Setup .pc file to allow use for cross-compiling

micmac1:
The Makefile is currently creating the pkg-config file using static lib
and include dir statements. Change that so that projects that
cross-compile hiredis can use pkg-config to setup other programs
depending on it.

Note: these projects (like OpenWrt) call pkg-config with arguments to
overwrite some variables in the .pc file, namely:

--define-variable=prefix=<...>
--define-variable=exec_prefix=<...>

Signed-off-by: Sebastian Kemper <sebastian_ml@gmx.net>


micmac1:
Hello,

I believe this was already addressed in master. Closing this. Thanks!

Kind regards,
Seb

PR #643, opened 2019-01-22
CMake Support

### LINKED ISSUE #611: Add CMake support
CMake would live side-by-side a simpler Makefile, and would allow easier integration into existing CMake projects, as well as easy building for Windows. Right now our Windows build environment seems to use Cygwin and/or Mingw32, which is less than optimal, especially if we are targeting MSVC for PHP/Python.

justinbrewer:
I've taken the liberty of moving @mnunberg's cmake branch towards feature parity with the current Makefile.

Supports:
 * Building
 * Installing
 * Running unit tests
 * Generating & installing pkg-config file

Missing:
 * Running unit tests through Valgrind
 * Building certain examples

Original Makefile is dropped. MSVC support, if desired, should be done in a subsequent PR.

Fixes #611. Fixes #641.


mnunberg:
I'm going to revert the change of removing the Makefile and then merge this

mnunberg:
I've merged this PR, sans the removal of the original Makefile

PR #644, opened 2019-01-28
Fix Invalid argument after redisAsyncConnectUnix

codehz:
related code:

```
int redisCheckConnectDone(redisContext *c, int *completed) {
    int rc = connect(c->fd, (const struct sockaddr *)c->saddr, c->addrlen);
```


mnunberg:
Thank you!

PR #645, opened 2019-02-20
SSL Support

mnunberg:
This PR adds SSL support, but also a rehaul of option parsing, as well as the ability to use timeouts in async mode.

### DIFF
--- a/adapters/libevent.h
+++ b/adapters/libevent.h
@@ -98,7 +98,7 @@ static void redisLibeventUpdate(void *privdata, short flag, int isRemove) {
     }
 
     event_del(e->ev);
-    event_assign(e->ev, e->base, e->context->c.fd, e->flags,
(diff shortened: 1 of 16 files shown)


adapters/libevent.h
```diff
@@ -34,48 +34,113 @@
...
+    short flags;
+    short state;
 } redisLibeventEvents;
 
-static void redisLibeventReadEvent(int fd, short event, void *arg) {
-    ((void)fd); ((void)event);
-    redisLibeventEvents *e = (redisLibeventEvents*)arg;
-    redisAsyncHandleRead(e->context);
+static void redisLibeventDestroy(redisLibeventEvents *e) {
+    free(e);
```
inbaryuval: set data in ac to NULL (e->context->ev.data = NULL)
1. it is safer
2. once we would add libeventDetach it would be very important (Redislabs needs it)

Issue #599, opened 2018-07-03
New hiredis fails to compile on Windows


galuszkak:
I started working on upgrade of hiredis in hiredis-py to new version of hiredis (from `master` branch) and while it compiles well on Linux & macOS then I have a problem on Windows. Stack trace from compilation:
```
c:\projects\hiredis-py\vendor\hiredis\sds.h(46) : error C2061: syntax error : identifier 'sdshdr5'
c:\projects\hiredis-py\vendor\hiredis\sds.h(46) : error C2059: syntax error : ';'
c:\projects\hiredis-py\vendor\hiredis\sds.h(46) : error C2449: found '{' at file scope (missing function header?)
c:\projects\hiredis-py\vendor\hiredis\sds.h(49) : error C2059: syntax error : '}'
c:\projects\hiredis-py\vendor\hiredis\sds.h(56) : error C2061: syntax error : identifier 'sdshdr16'
c:\projects\hiredis-py\vendor\hiredis\sds.h(56) : error C2059: syntax error : ';'
c:\projects\hiredis-py\vendor\hiredis\sds.h(56) : error C2449: found '{' at file scope (missing function header?)
c:\projects\hiredis-py\vendor\hiredis\sds.h(61) : error C2059: syntax error : '}'
c:\projects\hiredis-py\vendor\hiredis\sds.h(68) : error C2061: syntax error : identifier 'sdshdr64'
c:\projects\hiredis-py\vendor\hiredis\sds.h(68) : error C2059: syntax error : ';'
c:\projects\hiredis-py\vendor\hiredis\sds.h(68) : error C2449: found '{' at file scope (missing function header?)
c:\projects\hiredis-py\vendor\hiredis\sds.h(73) : error C2059: syntax error : '}'
c:\projects\hiredis-py\vendor\hiredis\sds.h(103) : error C2054: expected '(' to follow 'inline'
c:\projects\hiredis-py\vendor\hiredis\sds.h(103) : error C2085: 'sdsavail' : not in formal parameter list
c:\projects\hiredis-py\vendor\hiredis\sds.h(103) : error C2143: syntax error : missing ';' before '{'
c:\projects\hiredis-py\vendor\hiredis\sds.h(129) : error C2054: expected '(' to follow 'inline'
c:\projects\hiredis-py\vendor\hiredis\sds.h(129) : error C2085: 'sdssetlen' : not in formal parameter list
[...]

PR #652, opened 2019-03-31
Add MinGW support

mbitsnbites:
This PR introduces preliminary Windows support via MinGW.

The functionality is added in steps, so it is recommended to review the changes commit by commit.

The Windows/MinGW-specific changes are done in the "Port network layer to Winsock"-commit, which adds a simple socket compatibility layer (`sockcompat`), that essentially:
* Uses platform specific socket file descriptor types (`int` for POSIX, `SOCKET` for Windows).
* Includes platform specific headers (regular Unix/socket headers for POSIX, and winsock for Windows).
* Replaces selected POSIX socket API calls with Win32 wrappers that:
  * Handle type/signature conversions.
  * Transparently call `WSAStartup()` (once).
  * Update `errno` with the return value of `WSAGetLastError()` upon errors (so that the hiredis code can continue to use `errno` for its logic).

Still missing is a MinGW builder for travis or appveyor, but the branch builds under Ubuntu Linux using CMake + MinGW cross-compilation like so:

```bash
$ CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_BUILD_WITH_INSTALL_RPATH=on bath/to/source
$ ninja
```


mbitsnbites:
I didn't know what copyright owner/notice to use in the new source files `sockcompat.c` & `sockcompat.h`, so I just wrote my name there. I'll gladly replace it with whatever copyright you'd like (i.e. I have no interest in being the copyright owner).

mnunberg:
This looks very nice! I'm heading to redisconf tomorrow so I don't know if I'll have time to look at the TODO (ENOENT on ioctlsocket) this week, but I'd like to get that resolved before merging.

mbitsnbites:
I'll give it another try soon. I also noticed that I forgot to wrap `connect()` (which mostly means that errno may be wrong if connect fails), so I'll fix that too.

mbitsnbites:
Current findings: Using `connect()` on a non-blocking socket results in `WSAGetLastError() == WSAEWOULDBLOCK` [under Windows](https://docs.microsoft.com/en-us/windows/desktop/api/winsock2/nf-winsock2-connect). Since the socket always starts out as non-blocking in `net.c`, this is a problem. I'll continue investigating.

mbitsnbites:
I fixed the problem and removed the TODO: When `connect()` on non-blocking sockets results in `EWOULDBLOCK` on Windows, it is the same thing as `EINPROGRESS` on POSIX.

I did a special case conversion in the Win32 wrapper function for `connect()` (and I also rewrote the sockcompat layer to get more correct `errno` handling on Windows).

giorgiozoppi:
Interesting work. Do you know which actions for supporing MSVC17 or clang on windows?

mnunberg:
It would be awesome if we could have appveyor or whatever test this in CI. I can't guarantee that a future change won't just end up breaking this again.

mbitsnbites:
@giorgiozoppi I don't know, but it should not be prohibitively difficult. I think the gist of the work is to solve #599 (e.g. struct packing in the sds lib). I see it as a separate task.

mbitsnbites:
@mnunberg I agree. I took a quick look at your Travis file, but the way it is structured (as a matrix of compiler/OS/bitwidth), adding MinGW didn't feel straight forward (e.g. what does 32-bit MinGW on macOS mean?), so I think that it needs to be refactored.

The way I do it in BuildCache is this: https://github.com/mbitsnbites/buildcache/blob/master/.travis.yml#L39

I'm not sure how it would fit in your current Travis setup.

PR #662, opened 2019-05-05
The setsockopt and getsockopt API diffs from BSD socket and WSA one

dragonation:
The optval of `SO_RCVTIMEO` and `SO_SNDTIMEO` is a pointer to `DWORD`, not a pointer to `struct timeval` in Windows SDK.

ref:
* [getsockopt on MSDN](https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-getsockopt)
* [setsockopt on MSDN](https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-setsockopt)

### DIFF
--- a/sockcompat.c
+++ b/sockcompat.c
@@ -197,7 +197,7 @@ int win32_getsockopt(SOCKET sockfd, int level, int optname, void *optval, sockle
             socklen_t dwlen = 0;
             ret = getsockopt(sockfd, level, optname, (char *)&timeout, &dwlen);
             tv->tv_sec = timeout / 1000;
-            tv->tv_usec = timeout * 1000;
+            tv->tv_usec = (timeout * 1000) % 1000000;
         } else {
             ret = WSAEFAULT;
         }
--- a/sockcompat.c
+++ b/sockcompat.c
@@ -189,13 +189,31 @@ int win32_connect(SOCKET sockfd, const struct sockaddr *addr, socklen_t addrlen)
 }
 
 int win32_getsockopt(SOCKET sockfd, int level, int optname, void *optval, socklen_t *optlen) {
-    int ret = getsockopt(sockfd, level, optname, (char*)optval, optlen);
+    int ret = 0;
+    if ((level == SOL_SOCKET) && ((optname == SO_RCVTIMEO) || (optname == SO_SNDTIMEO))) {
+        struct timeval *tv = (struct timeval *)optval;
+        DWORD timeout = 0;
+        socklen_t dwlen = 0;
+        ret = getsockopt(sockfd, level, optname, (char *)timeout, &dwlen);
+        tv->tv_sec = timeout / 1000;
+        tv->tv_usec = timeout * 1000;
+        *optlen = sizeof (struct timeval);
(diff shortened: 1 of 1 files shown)


mnunberg:
This looks fine to me, though we don't ever use `getsockopt` in hiredis code

dragonation:
> This looks fine to me, though we don't ever use `getsockopt` in hiredis code

But we use setsockopt, which will cause the real timeout divided by 1000 in windows, if we connect redis server with a specified timeout, and usually it will make the connection very unstable when testing the new windows port of hiredis.

sockcompat.c
```diff
@@ -189,13 +189,35 @@ int win32_connect(SOCKET sockfd, const struct sockaddr *addr, socklen_t addrlen)
...
+    }
     _updateErrno(ret != SOCKET_ERROR);
     return ret != SOCKET_ERROR ? ret : -1;
 }
 
 int win32_setsockopt(SOCKET sockfd, int level, int optname, const void *optval, socklen_t optlen) {
-    int ret = setsockopt(sockfd, level, optname, (const char*)optval, optlen);
+    int ret = 0;
+    if ((level == SOL_SOCKET) && ((optname == SO_RCVTIMEO) || (optname == SO_SNDTIMEO))) {
+        struct timeval *tv = (struct timeval *)optval;
```
mnunberg: const struct timeval *tv = optval; eliminate casting plz

sockcompat.c
```diff
@@ -189,13 +189,35 @@ int win32_connect(SOCKET sockfd, const struct sockaddr *addr, socklen_t addrlen)
 }
 
 int win32_getsockopt(SOCKET sockfd, int level, int optname, void *optval, socklen_t *optlen) {
-    int ret = getsockopt(sockfd, level, optname, (char*)optval, optlen);
+    int ret = 0;
+    if ((level == SOL_SOCKET) && ((optname == SO_RCVTIMEO) || (optname == SO_SNDTIMEO))) {
+        if (*optlen >= sizeof (struct timeval)) {
+            struct timeval *tv = (struct timeval *)optval;
```
michael-grunder: Shouldn't need a cast here either.

mbitsnbites:
I suggest that #663 is merged before this PR since it adds Travis builders for MinGW and MSVC.

If this branch is rebased on master after the merge of #663, we get CI checking of the Windows code.

mbitsnbites:
> This looks fine to me, though we don't ever use `getsockopt` in hiredis code

Actually, it's used here: https://github.com/redis/hiredis/blob/master/net.c#L294, although only for `SOL_SOCKET, SO_ERROR`. However I think it's wise to have a correct implementation in `sockcompat.c` in case someone decides to use it later.

sockcompat.c
dragonation: Did it

sockcompat.c
dragonation: Did it

michael-grunder:
Is this safe to merge?  A few people have been having timeout related issues in Windows.

PR #663, opened 2019-05-08
Windows: MinGW fixes and Windows Travis builders

mbitsnbites:
Unfortunately the MSVC fixes in #658 broke MinGW builds (mainly due to using `_WIN32` instead of `_MSC_VER` when checking if the compiler is MS Visual Studio).

The ifdef-logic has been cleaned up slightly to support both MSVC and MinGW (and possibly Clang etc too).

In addition, a couple of Travis build steps are added that should catch future breakage of the MinGW and MSVC support.

### DIFF
--- a/sds.h
+++ b/sds.h
@@ -34,7 +34,7 @@
 #define __SDS_H
 
 #define SDS_MAX_PREALLOC (1024*1024)
-#ifdef _WIN32
+#ifdef _MSC_VER
 #define __attribute__(x)
 #endif
 
--- a/sds.h
+++ b/sds.h
@@ -135,20 +135,20 @@ static inline void sdssetlen(sds s, size_t newlen) {
         case SDS_TYPE_5:
             {
                 unsigned char *fp = ((unsigned char*)s)-1;
-                *fp = SDS_TYPE_5 | (newlen << SDS_TYPE_BITS);
+                *fp = (unsigned char)(SDS_TYPE_5 | (newlen << SDS_TYPE_BITS));
             }
             break;
         case SDS_TYPE_8:
-            SDS_HDR(8,s)->len = newlen;
+            SDS_HDR(8,s)->len = (uint8_t)newlen;
             break;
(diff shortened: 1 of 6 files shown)


mbitsnbites:
Apologies for being impatient, but any chance of this PR getting reviewed anytime soon? It would be good to get it into master to get better Travis checks for Windows PR:s. @mnunberg ?

sockcompat.h
```diff
@@ -50,7 +50,9 @@
 #include <ws2tcpip.h>
 #include <stddef.h>
 
+#ifdef _MSC_VER
 typedef signed long ssize_t;
```
mnunberg: What about win64?

sds.h
```diff
@@ -34,7 +34,7 @@
 #define __SDS_H
```
mnunberg: I think we pull sds directly from redis upstream.. so this might get overwritten at a later point in time.

sockcompat.h
mbitsnbites: How so? `_MSC_VER` is defined for all versions of the Microsoft Visual C++ compiler (and `_WIN32` is defined for both 32-bit and 64-bit Windows).

sds.h
mbitsnbites: AFAICT the source for sds has already diverged from the upstream version (due to previous hiredis patches). This patch merely fixes regressions due to earlier patches.

Going forward it would be good to sync with/upstream to redis.

mbitsnbites:
I'll try to rebase the branch and fix any conflicts...

mbitsnbites:
@mnunberg I've rebased and all Travis-CI checks passed.

sockcompat.h
mbitsnbites: By the way: Both the new travis builders (MinGW and MSVC) do 64-bit builds.

mnunberg: nvm, i forgot that `long` becomes 64 bits on 64 bit platforms.

PR #664, opened 2019-05-13
gtest for hiredis.c & async.c (not including SSL)

ashtul:
Colored out some sds but it runs when hard coded.


gkorland:
@mnunberg

cpptests/common.cpp
```diff
@@ -5,6 +5,20 @@ hiredis::ClientSettings hiredis::settings_g;
 
 using namespace hiredis;
 
+
+
+ClientSettings::ClientSettings(std::string connectType_, std::string str_) {
+    if(connectType_ == "env") {
```
mnunberg: Formatting

mnunberg: You might also want to normalize the casing here

cpptests/common.cpp
```diff
@@ -92,7 +106,7 @@ void ClientError::throwCode(int code) {
 }
 
 void ClientError::throwContext(const redisContext *c) {
-    const char *s;
+//    const char *what;
```
mnunberg: delete

cpptests/common.h
```diff
@@ -73,14 +76,14 @@ class ConnectError : public ClientError {
...
-    IOError(const char *s) : ClientError(s) {}
+    IOError() : ClientError() {}
+    IOError(const char *what) : ClientError(what) {}
 };
 
 class TimeoutError : public ClientError {
 public:
     TimeoutError() : ClientError("timed out") {}
-    TimeoutError(const char *s) : ClientError(s) {}
+    TimeoutError(const char *what) : ClientError(what) {}
```
mnunberg: Why the renaming? I mean, it might be a bit more intuitive...

cpptests/common.h
```diff
@@ -101,5 +104,9 @@ class CommandError : public ClientError {
     std::string errstr;
 };
 
+inline redisReply *castReply(void *reply) {
```
mnunberg: No need for the `inline` keyword

cpptests/main.cpp
```diff
@@ -28,7 +28,7 @@ int main(int argc, char **argv) {
     settings->m_ssl_key_path = HIREDIS_TEST_SSL_KEY;
 #endif
 
-    for (size_t ii = 1; ii < argc; ++ii) {
+    for (int ii = 1; ii < argc; ++ii) {
```
mnunberg: Why the change in type? did the compiler warn about this?

cpptests/t_basic.cpp
```diff
@@ -2,11 +2,32 @@
...
+    redisReader *reader;
+    void *reply = nullptr;;
+    int ret;
+    int i;    
+    
+    virtual void SetUp() {
+        reader = redisReaderCreate();
+    }
+    virtual void TearDown() {
+        if(reply == nullptr) { freeReplyObject(reply); }
```
mnunberg: != nullptr...

Or maybe better yet, provide a C++ wrapper class around Reply? (not necessary, but a good exercise)

cpptests/t_basic.cpp
```diff
@@ -19,6 +40,23 @@ static std::string formatCommand(const char *fmt, ...) {
     return xs;
 }
 
+#define INTEGER_WIDTH_TEST(fmt, type) do {                                                \
+    type value = 123;                                                                     \
+    len = redisFormatCommand(&cmd,"key:%08" fmt " str:%s", value, "hello");               \
+    ASSERT_TRUE(strncmp(cmd,"*2\r\n$12\r\nkey:00000123\r\n$9\r\nstr:hello\r\n",len) == 0 && \
```
mnunberg: ASSERT_STREQ or ASSERT_STREQ

cpptests/t_basic.cpp
```diff
@@ -19,6 +40,23 @@ static std::string formatCommand(const char *fmt, ...) {
...
+        len == 4+5+(12+2)+4+(9+2)) << "(" << #type << ")";                                \
+    free(cmd);                                                                            \
+} while(0)
+
+#define FLOAT_WIDTH_TEST(type) do {                                                       \
+    type value = 123.0;                                                                   \
+    len = redisFormatCommand(&cmd,"key:%08.3f str:%s", value, "hello");                   \
+    ASSERT_TRUE(strncmp(cmd,"*2\r\n$12\r\nkey:0123.000\r\n$9\r\nstr:hello\r\n",len) == 0 && \
```
mnunberg: ASSERT_STREQ...

cpptests/t_basic.cpp
```diff
@@ -31,4 +69,215 @@ TEST_F(FormatterTest, testFormatCommands) {
...
+    expected =  "*3\r\n$3\r\nSET\r\n$1\r\n%\r\n$1\r\n%\r\n";
+    ASSERT_EQ(expected, formatCommand("SET %% %%"))
+        << "literal %%";
+}
+
+TEST_F(FormatterTest, testTypeWidth) {
+    char *cmd;
+    int len;    
+
+    INTEGER_WIDTH_TEST("d", int);
```
mnunberg: It would be better if each test would have a different function, e.g.

```
TEST_F(FormatterTest, testIntWidthChar)...
```

Or, if you want to get really fancy, I know that GoogleTest has a combinatorial test functionality, see `TEST_P`

cpptests/t_basic.cpp
```diff
@@ -31,4 +69,215 @@ TEST_F(FormatterTest, testFormatCommands) {
...
+        len == 4+4+(3+2)+4+(3+2)+4+(3+2)) << "passing argc/argv without lengths";
+    free(cmd);
+
+    len = redisFormatCommandArgv(&cmd,argc,argv,lens);
+    ASSERT_TRUE(strncmp(cmd,"*3\r\n$3\r\nSET\r\n$7\r\nfoo\0xxx\r\n$3\r\nbar\r\n",len) == 0 &&
+        len == 4+4+(3+2)+4+(7+2)+4+(3+2)) << "passing argc/argv with lengths";
+    free(cmd);
+/*
+    sds_cmd = sdsempty();
+    len = redisFormatSdsCommandArgv(&sds_cmd,argc,argv,NULL);
```
mnunberg: why commented?

cpptests/t_basic.cpp
```diff
@@ -31,4 +69,215 @@ TEST_F(FormatterTest, testFormatCommands) {
...
+    sdsfree(sds_cmd);
+
+    sds_cmd = sdsempty();
+    len = redisFormatSdsCommandArgv(&sds_cmd,argc,argv,lens);
+    ASSERT_TRUE(strncmp(sds_cmd,"*3\r\n$3\r\nSET\r\n$7\r\nfoo\0xxx\r\n$3\r\nbar\r\n",len) == 0 &&
+        len == 4+4+(3+2)+4+(7+2)+4+(3+2)) << "sds by passing argc/argv with length";
+    sdsfree(sds_cmd);    */
+}
+
+TEST_F(ReplyReaderTest, testErrorHandeling) {
```
mnunberg: handling, not handeling

cpptests/t_basic.cpp
```diff
@@ -31,4 +69,215 @@ TEST_F(FormatterTest, testFormatCommands) {
...
+    redisReaderFeed(reader,(char*)"@foo\r\n",6);
```
mnunberg: Maybe redisReaderFeed should accept a `const char*` instead of a char, and we wouldn't have this problem.

cpptests/t_basic.cpp
```diff
@@ -31,4 +69,215 @@ TEST_F(FormatterTest, testFormatCommands) {
...
+    ret = redisReaderGetReply(reader,NULL);
+    ASSERT_TRUE(ret == REDIS_ERR &&
```
mnunberg: It's better to do multiple asserts rather than one big one; this makes it easier to see what went wrong

cpptests/t_client.cpp
```diff
@@ -29,10 +31,18 @@ class Client {
         }
     }
 
+    Client& operator=(const Client &other_) {
```
mnunberg: This will cause a crash, since the internal `ctx` will be freed twice.

cpptests/t_client.cpp
```diff
@@ -1,20 +1,22 @@
...
+#include <cassert>
+#include <climits>
+#include <netdb.h>
+
 #include "hiredis.h"
 #include "common.h"
 
-
-using namespace hiredis;
+namespace hiredis {
```
mnunberg: I'm not a fan of using scopes in source files

cpptests/t_client.cpp
```diff
@@ -1,20 +1,22 @@
...
 
 class Client {
 public:
-    operator redisContext*() {
+    redisContext* operator*() {
```
mnunberg: Why?

cpptests/t_client.cpp
```diff
@@ -29,10 +31,18 @@ class Client {
...
+        return *this;
+    }
+
+    operator redisContext*() const { return ctx; }
+
     void secureConnection(const ClientSettings& settings) {
         if (redisSecureConnection(
                 ctx, settings.ssl_ca(), settings.ssl_cert(),
-                settings.ssl_key(), NULL) != REDIS_OK) {
+                settings.ssl_key(), NULL) != REDIS_OK)
```
mnunberg: Opening brace on same line as the conditional, please

cpptests/t_client.cpp
```diff
@@ -90,23 +101,242 @@ class Client {
...
+
+    len = redisFormatCommand(&cmd, "SET foo bar");
+    ASSERT_TRUE(redisAppendFormattedCommand(*c, cmd, len) == REDIS_OK);
+    assert(redisGetReply(c, (void**)&reply) == REDIS_OK);
+    free(cmd);
+    freeReplyObject(reply);
+}
+
+TEST_F(ClientTestTCP, testBlockingConnection) {
+    reply = castReply(redisCommand(*c,"PING"));
```
mnunberg: Use the C++ `Client` wrapper object

cpptests/t_client.cpp
```diff
@@ -90,23 +101,242 @@ class Client {
...
+protected:
+    virtual void SetUp() {        
+        c.flushdb();
+    }
+    virtual void TearDown() {
+    //    if(reply != nullptr) { freeReplyObject(reply); }
+        
+    }
+
+    ClientSettings cs;
```
mnunberg: This should be a global; if there needs to be overrides, it should be done on a per-function basis.

setgtest
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# cpptests gtest
+
+mkdir contrib
+./get_gtest.sh 
+cmake CMakeLists.txt -DENABLE_TESTS=1
```
mnunberg: -DCMAKE_BUILD_TYPE=DEBUG

cpptests/main.cpp
ashtul: Yes

cpptests/common.h
ashtul: Different names for different exception types

cpptests/t_client.cpp
ashtul: I ended up adding it later somewhere else. So now we have both!

cpptests/t_client.cpp
ashtul: I caught it later and fixed it.
Not sure why you see it this way.

cpptests/t_basic.cpp
ashtul: This one is truly for the master of CMAKE.
It would not compile since it can't link with sds.o as cmake doesn't put it in the library.

cpptests/t_basic.cpp
ashtul: Good call.
Not sure why was it there to begin with. It is indeed const char*

cpptests/t_client.cpp
ashtul: changed to memcpy

cpptests/t_client.cpp
ashtul: Why?

cpptests/t_client.cpp
ashtul: Why?

cpptests/t_basic.cpp
ashtul: true this

ashtul:
@mnunberg @gkorland 
Updated copy with async and PubSub

ashtul:
To run these tests first initilize with:
`./setgtest`
then run:
`coverage.sh`

PR #670, opened 2019-05-28
fix timeout code in windows

jman-krafton:
Timeout was not working in Windows.

### DIFF
--- a/net.c
+++ b/net.c
@@ -310,11 +310,18 @@ int redisCheckSocketError(redisContext *c) {
 }
 
 int redisContextSetTimeout(redisContext *c, const struct timeval tv) {
-    if (setsockopt(c->fd,SOL_SOCKET,SO_RCVTIMEO,&tv,sizeof(tv)) == -1) {
+    const void *to_ptr = &tv;
+    size_t to_sz = sizeof(tv);
+#ifdef _WIN32
+    DWORD timeout_msec = tv.tv_sec * 1000 + tv.tv_usec / 1000;
+    to_ptr = &timeout_msec;
+    to_sz = sizeof(timeout_msec);
+#endif
+    if (setsockopt(c->fd,SOL_SOCKET,SO_RCVTIMEO,to_ptr,to_sz) == -1) {
         __redisSetErrorFromErrno(c,REDIS_ERR_IO,"setsockopt(SO_RCVTIMEO)");
         return REDIS_ERR;
     }
-    if (setsockopt(c->fd,SOL_SOCKET,SO_SNDTIMEO,&tv,sizeof(tv)) == -1) {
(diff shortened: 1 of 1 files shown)


net.c
```diff
@@ -63,7 +63,15 @@ int redisNetRead(redisContext *c, char *buf, size_t bufcap) {
         if ((errno == EWOULDBLOCK && !(c->flags & REDIS_BLOCK)) || (errno == EINTR)) {
             /* Try again later */
             return 0;
-        } else {
+        }
+		#ifdef _WIN32
+		else if(errno == ETIMEDOUT && (c->flags & REDIS_BLOCK))
+		{	/* especially in windows */
```
mnunberg: Can you fix the formatting here? (spaces, no tabs; opening brace on same line, etc.)

mnunberg: I also don't see how this block is specific to Windows.

net.c
```diff
@@ -63,7 +63,15 @@ int redisNetRead(redisContext *c, char *buf, size_t bufcap) {
         if ((errno == EWOULDBLOCK && !(c->flags & REDIS_BLOCK)) || (errno == EINTR)) {
             /* Try again later */
             return 0;
-        } else {
+        }
+        #ifdef _WIN32
```
mnunberg: This doesn't need an `#ifdef`

mnunberg: also, fix brace on same line

net.c
jman-krafton: > I also don't see how this block is specific to Windows.

In fact, I think so. I added the block in order to clarify a timeout error code when giving the timeout value on blocking mode. But, originally the block wasn't there and I was uncertain that I coded correct, so I specified the block to only Windows.

If this block doesn't have any problem, I can remove #preprocessor or else I will remove the block.

net.c
jman-krafton: > also, fix brace on same line

i fixed it.

jman-krafton: > This doesn't need an `#ifdef`

i removed #ifdef ~ #endif

mnunberg: You're still using 8 space indentation.. are you using a proper editor? :)

mnunberg:
Oh, and when you're done, please squash into two commits

net.c
jman-krafton: My default indent setting(VS2017) is a tab. :]
I (really) have fixed all formatting mistakes.

jman-krafton:
> Oh, and when you're done, please squash into two commits
I am done.

HelloAWorld:
I aslo got timeout problem on windows, and I commit issues 
[redis timeout issues #677](https://github.com/redis/hiredis/issues/677)
Please review it if it's the same problem.

PR #671, opened 2019-05-30
redisReaderGetReply leak memory

movebean:
I think when the parameter `reply` is NULL, then only set r->reply to NULL at the tail of function will lead to memory leak, so I add some code to call freeObject to free r->reply under some conditions of course.

### DIFF
--- a/read.c
+++ b/read.c
@@ -590,8 +590,11 @@ int redisReaderGetReply(redisReader *r, void **reply) {
 
     /* Emit a reply when there is one. */
     if (r->ridx == -1) {
-        if (reply != NULL)
+        if (reply != NULL) {
             *reply = r->reply;
+        } else if (r->reply != NULL && r->fn && r->fn->freeObject) {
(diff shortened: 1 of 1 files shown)


read.c
```diff
@@ -590,8 +590,11 @@ int redisReaderGetReply(redisReader *r, void **reply) {
 
     /* Emit a reply when there is one. */
     if (r->ridx == -1) {
-        if (reply != NULL)
+        if (reply != NULL) {
             *reply = r->reply;
+        } else if (r->reply != NULL && r->fn && r->fn->freeObject) {
+            r->fn->freeObject(r->reply);
+        }
```
ifduyue: This is a dead branch, program will never go into here.

movebean: If I call the function like redisReaderGetReply(r, NULL); ?  Is it a dead branch?

ifduyue: @movebean I see the difference. you're right.

Issue #646, opened 2019-02-23
Allow to change SSL Mode


huyphams:
Updated:
- Hiredis SSL default mode is SSL_VERIFY_PEER, which requires a valid certification.
- Amazon ElasticCache (and some servers) doesn't require a certification because the client is a trusted source. So it would be great if Hiredis support an ability to change the SSL mode in the `redisSecureConnection`

======
Original content:

I can connect to the ElastiCache instance without SSL successfully but when I tried to connect to the ElastiCache using SSL, I got this error:

I'm using the master with this merge: https://github.com/redis/hiredis/pull/645 thanks for supporting SSL

```
ST(0x10). before/connect initialization. R(0x1)U
ST(0x1001). before/connect initialization. R(0x1)U
ST(0x1001). SSLv2/v3 write client hello A. R(0x1)U
ST(0x1001). unknown state. R(0x1)U
ST(0x4008). unknown state. R(0x230)F
ST(0x1002). error. R(0xffffffff)U
ST(0x1002). error. R(0xffffffff)U
```

Error message: SSL_connect() failed

ElastiCache Redis 5.0.0
Encryption in-transit: Yes
Hiredis at commit: https://github.com/redis/hiredis/commit/297de989a647a6f96f25aabea267aa0d18c43328

Let me know if you need anything else, thanks!

mnunberg:
have you tried with openssl s_client? this might help debug some issues beforehand.

huyphams:
I've tried the lasted commit on master but still got the same error, here is the openssl s_client detail.
```
[centos]$ openssl s_client -connect master.xxxx.ewbgtg.use1.cache.amazonaws.com:6379
CONNECTED(00000003)
depth=4 C = US, O = "Starfield Technologies, Inc.", OU = Starfield Class 2 Certification Authority
verify return:1
depth=3 C = US, ST = Arizona, L = Scottsdale, O = "Starfield Technologies, Inc.", CN = Starfield Services Root Certificate Authority - G2
verify return:1
depth=2 C = US, O = Amazon, CN = Amazon Root CA 1
verify return:1
depth=1 C = US, O = Amazon, OU = Server CA 1B, CN = Amazon
verify return:1
depth=0 CN = *.xxxx.ewbgtg.use1.cache.amazonaws.com
verify return:1
---
Certificate chain
 0 s:/CN=*.xxxx.ewbgtg.use1.cache.amazonaws.com
[...]

PR #672, opened 2019-06-01
Support changing SSL Mode

huyphams:
- Hiredis SSL default mode is SSL_VERIFY_PEER, which requires a valid certification.
- Amazon ElasticCache (and some servers) doesn't require a certification because the client is a trusted source. So it would be great if Hiredis support an ability to change the SSL mode in the redisSecureConnection.
- It's also a good thing, we should clarify the SSL mode before making the connection.

Original issue: https://github.com/redis/hiredis/issues/646

### DIFF
--- a/hiredis.h
+++ b/hiredis.h
@@ -235,11 +235,24 @@ redisContext *redisConnectUnixNonBlock(const char *path);
 redisContext *redisConnectFd(redisFD fd);
 
 /**
+ *  Redis SSL options
+ */
+# define HIREDIS_SSL_VERIFY_NONE 0x00
+# define HIREDIS_SSL_VERIFY_PEER 0x01
+
+/**
  * Secure the connection using SSL. This should be done before any command is
  * executed on the connection.
  */
 int redisSecureConnection(redisContext *c, const char *capath, const char *certpath,
                           const char *keypath, const char *servername);
+
+/**
+ * Secure the connection using SSL with options. This should be done before any command is
+ * executed on the connection.
+ */
+int redisSecureConnectionWithOptions(redisContext *c, const char *capath, const char *certpath,
+                          const char *keypath, const char *servername, int ssloptions);
 
 /**
  * Reconnect the given context using the saved information.
--- a/sslio.c
+++ b/sslio.c
@@ -88,7 +88,7 @@ void redisFreeSsl(redisSsl *ssl){
 }
 
 int redisSslCreate(redisContext *c, const char *capath, const char *certpath,
-                   const char *keypath, const char *servername) {
+                   const char *keypath, const char *servername, int ssloptions) {
     assert(!c->ssl);
     c->ssl = calloc(1, sizeof(*c->ssl));
     static int isInit = 0;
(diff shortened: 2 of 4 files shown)


hiredis.h
```diff
@@ -234,12 +234,22 @@ redisContext *redisConnectUnixWithTimeout(const char *path, const struct timeval
 redisContext *redisConnectUnixNonBlock(const char *path);
 redisContext *redisConnectFd(redisFD fd);
 
+/**
+ *  Redis SSL mode
+ */
+enum redisSslMode {
```
mestery: It may be nice to do a `typedef enum redisSslMode redisSslMode` here so you can then skip the enum when declaring these in other places.

sslio.c
```diff
@@ -87,8 +87,8 @@ void redisFreeSsl(redisSsl *ssl){
     free(ssl);
 }
 
-int redisSslCreate(redisContext *c, const char *capath, const char *certpath,
-                   const char *keypath, const char *servername) {
+int redisSslCreate(redisContext *c, const char *caPath, const char *certPath,
```
mestery: Any specific reason why these were changed to capitalize `Path` in each variable name?

mestery:
I've pulled this and tested it and can confirm it works as @huyphams indicates.

sslio.c
huyphams: I follow the camelCase conventions in the header file (https://github.com/redis/hiredis/blob/master/sslio.h) so I changed it to capitalize.

examples/example-libevent-ssl.c
```diff
@@ -50,14 +50,15 @@ int main (int argc, char **argv) {
     const char *cert = argv[4];
     const char *certKey = argv[5];
     const char *caCert = argc > 5 ? argv[6] : NULL;
+    enum redisSslMode sslMode = HIREDIS_SSL_VERIFY_PEER;
```
mnunberg: Use a typedef

hiredis.h
```diff
@@ -234,12 +234,22 @@ redisContext *redisConnectUnixWithTimeout(const char *path, const struct timeval
...
+    HIREDIS_SSL_VERIFY_NONE = 0,
+    HIREDIS_SSL_VERIFY_PEER,
+    HIREDIS_SSL_VERIFY_FAIL_IF_NO_PEER_CERT,
```
mnunberg: Hiredis is a client; only `VERIFY_PEER` and `VERIFY_NONE` make sense here.

hiredis.h
```diff
@@ -234,12 +234,22 @@ redisContext *redisConnectUnixWithTimeout(const char *path, const struct timeval
...
+    HIREDIS_SSL_VERIFY_CLIENT_ONCE
+};
+
 /**
  * Secure the connection using SSL. This should be done before any command is
  * executed on the connection.
  */
-int redisSecureConnection(redisContext *c, const char *capath, const char *certpath,
-                          const char *keypath, const char *servername);
+int redisSecureConnection(redisContext *c, const char *caPath, const char *certPath,
```
mnunberg: Please don't change the naming conventions for argument names. If you'll notice, this is the convention for argument names in other functions as well..

examples/example-libevent-ssl.c
mnunberg: I'd actually prefer that this was a bitmask of options (and therefore a `#define`); this way we can semantically extend it later on if we need to pass other flags.

Maybe we can even do this with a `redisSslOptions` struct for future use?

hiredis.h
huyphams: Sure, let me fix it.

mnunberg:
I actually think we are going to go with a more comprehensive option, such as redisSecureConnectionWithOptions(...) and have the SSL_CTX option as a possible input.

huyphams:
I've updated the code and separated it with the new function
redisSecureConnectionWithOptions

yossigo:
@huyphams Please have a look at the latest changes, you can now simply configure OpenSSL yourself and provide a `SSL*`, I believe it would be the most generic solution.

nysenthil:
This SSL support is a much needed feature for hiredis for a long time. I compiled the master branch code as it existed on Nov/20/2019 and tested it with a major Cloud based Redis service. It worked fine. Thanks to @mnunberg  and @yossigo for an elegant implementation. I like the flexibility of the redisInitiateSSL C function you added. I request you to create a hiredis release 0.15.0 of this code base with SSL support. Now that hiredis supports SSL, I will wait for a similar SSL support in the cpp-hired-cluster client library as well.

yossigo:
@nysenthil Thanks for your feedback, I think there is some more planned work towards an official release but it's definitely planned.

michael-grunder:
I believe this was solved in #821 as you can now manually configure your `SSL_CTX` however you like.

Issue #683, opened 2019-07-03
Remove an unnecessary null pointer check


elfring:
[An extra null pointer check is not needed](https://stackoverflow.com/questions/18775608/free-a-null-pointer-anyway-or-check-first "Free a null pointer anyway or check first?") in [the function “_redisContextConnectTcp”](https://github.com/redis/hiredis/blob/ca153fa4a8ae84021bdce7ec47e5339eef1d9212/net.c#L438 "Update candidate").

PR #684, opened 2019-07-03
Remove unnecessary null check before free

qlyoung:
Fixes: #683


elfring:
Thanks for your small source code improvement.

PR #697, opened 2019-08-05
Port RESP3 support from Redis.

yossigo:
This is a port of the RESP3 support added to the hiredis version bundled as part of Redis.

### DIFF
--- a/read.h
+++ b/read.h
@@ -54,6 +54,14 @@
 #define REDIS_REPLY_NIL 4
 #define REDIS_REPLY_STATUS 5
 #define REDIS_REPLY_ERROR 6
+#define REDIS_REPLY_DOUBLE 7
+#define REDIS_REPLY_BOOL 8
+#define REDIS_REPLY_VERB 9
+#define REDIS_REPLY_MAP 9
+#define REDIS_REPLY_SET 10
+#define REDIS_REPLY_ATTR 11
+#define REDIS_REPLY_PUSH 12
+#define REDIS_REPLY_BIGNUM 13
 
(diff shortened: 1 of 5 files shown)


read.h
```diff
@@ -54,6 +54,14 @@
 #define REDIS_REPLY_NIL 4
 #define REDIS_REPLY_STATUS 5
 #define REDIS_REPLY_ERROR 6
+#define REDIS_REPLY_DOUBLE 7
+#define REDIS_REPLY_BOOL 8
+#define REDIS_REPLY_VERB 9
```
michael-grunder: Do you know if `REDIS_REPLY_VERB` and `REDIS_REPLY_MAP` resolving to 9 intentional here?  Looks like it's in Redis proper too, but I don't see it referenced anywhere else.

michael-grunder:
I'm willing to create tests but I think we should probably merge this.

read.h
yossigo: @michael-grunder I'd guess it's a mistake in the original commit.

yossigo:
My goal here is to merge RESP3 changes made by @antirez back to mainline hiredis, so the latest (with SSL support) can be used in `deps`.  I guess it's really only work in progress towards full RESP3, but there's probably no harm in merging it incrementally.

mbitsnbites:
This merge broke MSVC builds.

These kind of problems could be prevented if PR branches are rebased on top of the latest master before merging. That way the tested code represents the post-merge state, and the latest builds & tests from master will be run on the PR.

PR #704, opened 2019-08-28
Fix some SSL related error handling.

yossigo:
* When SSL is not compiled, produce an explicit error.
* Fix error codes for SSL configuration errors.


yossigo:
This is no longer relevant after the SSL restructuring.

PR #707, opened 2019-08-29
SSL Code Reorganization.

yossigo:


yossigo:
Deleting this one was supposed to be a draft.

Issue #705, opened 2019-08-28
SSL Support Reorganization


yossigo:
Here are some thoughts about reorganizing SSL support in hiredis.

1. Add a `redisUseSSL(redisContext *, SSL*)` function to let callers supply a pre-configured `SSL*` context and thus have full control over OpenSSL configuration options.
2. The above will also imply that hiredis does not need to initialize OpenSSL global state, as `redisSecureConnection()` does now and may cause conflicts with other users of OpenSSL.
3. Reorganize the code so the SSL-related functions will not be compiled and won't be available if hiredis was compiled without SSL support.
4. As an extension to the above, make it possible to compile only SSL support as a separate library (with its own header file) which can be used as an extension.  This should make it easier to package hiredis and hiredis+ssl support separately if one wishes to do so (see Debian libevent/libevent-openssl as an example).

To achieve [3] and [4] I think `sslio.c` can become `ssl.c` and be self contained.  The use of the `REDIS_SSL` flag to determine which read/write functions should be used can be replaced with simple function callbacks, and so can the use of `struct redisSsl` so the rest of hiredis is completely decoupled from the SSL implementation.

@mnunberg does this make sense to you?

yossigo:
A bunch of function pointers yes, to facilitate a kind of run-time polymorphism on top of `redisContext`. I don't see why dlopen will be necessary though.

yossigo:
@mnunberg yes, but the way I see it you'll be able to build hiredis into `hiredis.so` (or a static lib of course) and `hiredis_ssl.so`.

If you never use SSL, just link with `-lhiredis` but you'll not have `redisSecureConnection()` etc.
If you want to use SSL, you need to link with `-lhiredis -lhiredis_ssl -lssl -lcrypto` and you'll have the SSL functions.

And as you indicate, by default we can just build a single library with/without SSL support. But thinking about how this can get packaged for Linux distros etc, I think the core + SSL support approach is a clear winner.

mnunberg:
Oh, I see, that makes sense.

yossigo:
@mnunberg I'm tempted to make this the option rather than complicate the Makefile and live in a world where hiredis.so could be with or without SSL support.  It basically means if you need SSL you always have an extra `#include "hiredis_ssl"` and `-lhiredis_ssl`, but then we core hiredis is never contaminated with SSL dependencies.  Thoughts about it?

mnunberg:
[...]

PR #708, opened 2019-08-29
SSL Reorganization

yossigo:
This is an implementation of the changes discussed in #705.


async.c
```diff
@@ -347,7 +314,7 @@ void redisAsyncFree(redisAsyncContext *ac) {
 }
 
 /* Helper function to make the disconnect happen and clean up. */
-static void __redisAsyncDisconnect(redisAsyncContext *ac) {
+void __redisAsyncDisconnect(redisAsyncContext *ac) {
```
mnunberg: I really hate the `__`:'( but alas this is not a concern in this PR

hiredis.h
```diff
@@ -193,8 +190,21 @@ typedef struct {
...
+    void (*free_privdata)(void *);
+    void (*async_read)(struct redisAsyncContext *);
+    void (*async_write)(struct redisAsyncContext *);
+    int (*read)(struct redisContext *, char *, size_t);
+    int (*write)(struct redisContext *);
+} redisContextFuncs;
+
 /* Context for a connection to Redis */
 typedef struct redisContext {
+    redisContextFuncs *funcs;   /* Function table */
```
mnunberg: const?

yossigo: Definitely, will add.

PR #720, opened 2019-10-07
Update README with SSL support.

yossigo:


michael-grunder:
Merged, thanks!

michael-grunder:
@yossigo You put `USE_TLS=1` in the READMD.md, but I shouldn't it actually be [USE_SSL=1](https://github.com/redis/hiredis/blob/42a2e679bebf3712a2d4403b58fc583dc90357e1/Makefile#L9)

yossigo:
@michael-grunder thanks, you are right of course.

PR #724, opened 2019-10-25
add static lib

animic:


langep:
What is blocking this from getting merged?

PR #727, opened 2019-11-10
Adding an option to DISABLE_TESTS

pbotros:
Useful when hiredis is used as a CMake dependency in other projects and
added via add_subdirectory(). With DISABLE_TESTS on, `make test` in the
parent project won't run hiredis tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ INCLUDE(GNUInstallDirs)
 PROJECT(hiredis)
 
 OPTION(ENABLE_SSL "Build hiredis_ssl for SSL support" OFF)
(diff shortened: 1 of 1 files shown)


pbotros:
@mnunberg ^ if you're the right person to tag - thanks!

pbotros:
@michael-grunder or @yossigo, could you PTAL? (sorry, tagging random people with recent commits)

michael-grunder:
This seems like a reasonable change to me, but I don't know CMake all that well.  That said, the option works as I would expect so I'll merge it tomorrow assuming there aren't any objections.

michael-grunder:
Merged, thanks!

pbotros:
Thanks @michael-grunder !

PR #734, opened 2019-11-24
Coverage of hiredis.c

ashtul:
In hiredis.c, some of the uncovered functions and lines are likely to be dead.


ashtul:
@yossigo 
This is a very small commit but...
A. It covers all of hiredis.c (not including blocking functions)
B. I would like to see the process of getting PRs in.

hiredis.c
```diff
@@ -122,10 +122,10 @@ static void *createStringObject(const redisReadTask *task, char *str, size_t len
         return NULL;
 
     buf = malloc(len+1);
-    if (buf == NULL) {
+    if (buf == NULL) {          // LCOV_EXCL_START
```
yossigo: @ashtul Did you consider the tradeoff between more accurate coverage and code clutter ?
Not sure it's worth having LCOV-specific tags all over the place.

test.c
```diff
@@ -452,13 +472,25 @@ static void test_reply_reader(void) {
     /* Regression test for issue #45 on GitHub. */
     test("Don't do empty allocation for empty multi bulk: ");
     reader = redisReaderCreate();
-    redisReaderFeed(reader,(char*)"*0\r\n",4);
+    redisReaderFeed(reader,(char*)"*0\r\n",8);
```
yossigo: I think this is wrong, the length of the string is 4 bytes.

hiredis.c
ashtul: I see your point
This is really your call
I will remove

ashtul:
@yossigo Shell we add CodeCov tag?

michael-grunder:
There's good stuff in here that we should probably get merged before v1.0.0

I went ahead and rebased your PR against the new pluggable-allocation branch to make eventual merging easier.

[Link to rebase](https://github.com/michael-grunder/hiredis/tree/gcov-rebase)

PR #736, opened 2019-11-25
with libuv async request, two problem and solutions on Windows.

xingyun86:


michael-grunder:
I cherry-picked the change to `uv_poll_init_socket` and merged it via #784.

I'm somewhat uncomfortable just slapping a `sleep(0)` call inside `_redisContextConnectTcp` though.

I will do more research but I can't find people discussing this fix anywhere.  If it is the "right" thing to do in Windows, I think we should put it in our win32 `setsockopt` compatibility function instead.

Thank you for the fix!

michael-grunder:
I've created a new issue to track the sleep discussion in #785 so we don't forget about it.

PR #737, opened 2019-11-25
Fix dead code in sslLogCallback relating to should_log variable.

natoscott:
Coverity scans found that the should_log logic in sslLogCallback
is not working as expected because the variable is not correctly
initialised (the conditional code before logging always sets the
value to 1, which it already is).


michael-grunder:
Merged, thanks!

Issue #740, opened 2019-12-11
use redisReaderGetReply(r, NULL) cause memory leak


coolheart77:
![image](https://user-images.githubusercontent.com/21165659/70619567-335b5e80-1c0d-11ea-90e8-e1e78961e1ae.png)
if reply==NULL， users don't care reply，this will cause memory leak

michael-grunder:
This fix has been merged, thanks for the report!

PR #741, opened 2019-12-12
Free the reply in redisGetReply when passed NULL

michael-grunder:
We currently perform a NULL check in redisGetReply and don't push the
reply back to the caller, but we don't free any reply meaning that this
will leak memory:

redisGetReply(context, NULL);

This change simply frees the reply if we were passed NULL.

Addresses #740


michael-grunder:
This code hasn't been changed in almost ten years but it seems like we should free the reply especially since we're already doing a NULL check.

cc: @yossigo

Issue #747, opened 2020-01-09
Null pointer dereferencing in dict.c and async.c


shouc:
The following code never aborts when `malloc` is unsuccessful, causing dereferencing of null pointers.

async.c:61
```C
redisCallback *dup = malloc(sizeof(*dup));
memcpy(dup,src,sizeof(*dup)); // dup may be null
return dup;
```

dict.c:75
```C
dict *ht = malloc(sizeof(*ht));
_dictInit(ht,type,privDataPtr); // ht may be null
```

dict.c:146
```C
entry = malloc(sizeof(*entry));
entry->next = ht->table[index]; // entry may be null
```

dict.c:261
```C
dictIterator *iter = malloc(sizeof(*iter));
iter->ht = ht; // iter may be null
```

lamby:
This has been given the CVE ID [CVE-2020-7105](https://security-tracker.debian.org/tracker/CVE-2020-7105).

michael-grunder:
We have an open [PR](https://github.com/redis/hiredis/pull/638) that has this [specific change](https://github.com/redis/hiredis/pull/638/files#diff-8bd6782b369f6afec2c090b37d553d46L58) in it along with a few other NULL pointer deref fixes although it does not immediately abort on failure to reallocate.

Just adding it here for reference.

shouc:
fixed in #752 & #638

lamby:
Shall we close #752 as well then? I'm a little lost, especially as #638 has a bunch of other changes that will be very difficult to backport to older, released, versions of hiredis.

michael-grunder:
Apologies, I wasn't trying to confuse the situation.  I just wanted to mention that #638 contains the same change (in addition to many more changes).

Merging #752 now and then #638 later seems reasonable to me.

shouc:
#752 contains more fixes on nullptr dereferencing so I also think merging these two is reasonable

PR #752, opened 2020-01-19
CVE-2020-7105: Abort if malloc() was unsuccessful

lamby:


lamby:
This is to address https://github.com/redis/hiredis/issues/747

mnunberg:
Is there a functional advantage of crashing via abort vs crashing via segfault?

lamby:
The segfault might be exploitable, and the `abort` is more debuggable outside of a security context.

shouc:
I think a better way to fix this is to abstract out the handling function to header files. For example:
`#define OOM_HANDLER abort()` and use OOM_HANDLER to substitute abort(), which makes upstream developer easier to define their own try catch logic.

lamby:
Ah, good idea. I was not aware of that particular idiom, thanks.

lamby:
Something like this? :)

shouc:
lgtm but I am not the owner

mnunberg:
Perhaps just wrap the entire allocation in a macro/function?

lamby:
> Perhaps just wrap the entire allocation in a macro/function?

This could work. But:

* Is there a Don't Repeat Yourself version of doing this that can support `malloc` and `calloc` (used at least 9 times)
* It would be a fairly invasive code change (fine for `master`) but perhaps rather unwieldy to backport.

mnunberg:
1. just wrap calloc, malloc, realloc, strdup, etc etc. I mean, this is what redis server does itself.
2. I personally don't think that a segfault on NULL is all that big a deal, but I think we should simply start first at those areas which have been identified. As far as the invasiveness, a one-line change in a  bunch of places is going to be far less invasive than a multi-line change in those same areas.

michael-grunder:
> just wrap calloc, malloc, realloc, strdup, etc etc. I mean, this is what redis server does itself

That's what I would suggest as well.  Given that we often _do_ handle a `NULL` reply (e.g. returning REDIS_ERROR_OOM) we should probably name them something else. 

What about `hr_safe_malloc`, `hr_safe_realloc`, `hr_safe_calloc`, and `hr_safe_strdup`?

I quickly sketched out a POC [over on my fork](https://github.com/michael-grunder/hiredis/pull/new/safe-alloc).

If we're all good with that direction I'm happy to formalize (and actually test) the changes.

lamby:
> I quickly sketched out a POC over on my fork.

FYI this links to the "New PR" page and not the PR itself. Did you forget to push it? I [don't see the branch in your fork..](https://github.com/michael-grunder/redis/tree/safe-alloc). :)

michael-grunder:
Apparently I can't link things properly :smile:

[link to my POC branch](https://github.com/michael-grunder/hiredis/commits/safe-alloc)
[link to the diff](https://github.com/michael-grunder/hiredis/compare/master..safe-alloc)

lamby:
Thanks; nice work. I suppose the other approach might be to emulate what Redis' `src/zmalloc.c` does by `#define realloc(ptr,size) foo_realloc(ptr,size)`  etc.?

michael-grunder:
I like that strategy as well, but we probably need two versions because there are a bunch of places where OOM is handled in a non-fatal way.

@mnunberg let us know if this strategy seems reasonable.

michael-grunder:
As an alternative what do people think about getting #638 merged first?

That has some good stuff in it in addition to the malloc NULL checks.  In particular the the `redisAsync*Append` functions.

shouc:
lgtm, the only conflict is in `async.c`, where #638 uses following approach:
```c
if (NULL != (dup = malloc(sizeof(*dup))))
        memcpy(dup, src, sizeof(*dup));
```

lamby:
> what do people think about getting #638 merged first?

I'd quite very like this one (or something similar) merged first as I am trying to ensure I can cherry-pick a minimal patch for older and already-shipped versions of hiredis, hence trying to reduce the number of changes. It would be difficult for me to ship out #638's wide-reaching changes in a safe and confident manner (and might even be difficult to backport in a pure code sesne as it changes quite a lot).

mnunberg:
Re #638 I'm not a fan of mindlessly checking for NULL so much and so explicitly that it makes the actual code hard to read. I think that PR has too much boilerplate, though I do appreciate what he's trying to do over there.

@michael-grunder I'm in favor of a prefixed malloc, maybe `hiredis_malloc` or whatever... or just `hi_malloc`

michael-grunder:
@mnunberg Seems reasonable to me.  I'll package it up and open a PR for review.

michael-grunder:
Closing in favor of #754, please redirect all bikeshedding over there.

stevelipinski:
Will a new release/tag be issued (0.15.0?) since this addresses a CVE?

PR #754, opened 2020-01-23
Safe allocation wrappers

michael-grunder:
Adds allocation wrappers that invokes a defined OOM handler.

My goal was the smallest surface area of changes possible.  Once we merge this it might be prudent to replace every `malloc`/`free` call with our wrappers.


lamby:
Just quoting the [last parts of the conversation in a merged PR](https://github.com/redis/hiredis/pull/752#issuecomment-577429138) from @michael-grunder 

>  @lamby I totally understand not wanting such a huge change just for a patch.

Thanks... and I think the changes in this PR would work for me. :+1:

lamby:
Would be great to get some more eyes on this. :)

michael-grunder:
> are we using asprintf and friends anywhere?

No, but `sds` has something similar (e.g. [sdscatfmt](https://github.com/redis/hiredis/blob/0501c623c91344e54cb2775a91509650960789b1/sds.c#L582)) although it appears that we're testing for NULL every time.

michael-grunder:
Merged, thanks everyone!

alloc.c
```diff
@@ -0,0 +1,65 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "fmacros.h"
+#include "alloc.h"
```
peycho: Isn't that deprecated?

peycho:
```
In file included from src/redisq_handler.h:5:0,
                 from src/am.cpp:14:
/usr/local/include/hiredis/hiredis.h:45:10: fatal error: alloc.h: No such file or directory
 #include "alloc.h" /* for allocation wrappers */
          ^~~~~~~~~
compilation terminated.
In file included from src/redisq_handler.h:5:0,
                 from src/redisq_handler.cpp:7:
/usr/local/include/hiredis/hiredis.h:45:10: fatal error: alloc.h: No such file or directory
 #include "alloc.h" /* for allocation wrappers */
          ^~~~~~~~~
compilation terminated.
Makefile:58: recipe for target 'am' failed
make: *** [am] Error 1
```

michael-grunder: Ahh, are you worried about a name conflict with the old "alloc.h" file?

peycho: Yup. I though that was the problem at first. Anyway, you guys fixed the problem very fast. Thanks.

async.c
```diff
@@ -57,7 +58,7 @@ static unsigned int callbackHash(const void *key) {
 
 static void *callbackValDup(void *privdata, const void *src) {
     ((void) privdata);
-    redisCallback *dup = malloc(sizeof(*dup));
+    redisCallback *dup = hi_malloc(sizeof(*dup));
```
oshamash, line 61: Can still result in a SEGFAULT,
If hi_malloc returns NULL -> next call to memcpy will SEGFAULT

lamby: Just waking up here but how can hi_malloc return NULL if `HIREDIS_OOM_HANDLER` is called? (eg. `_exit`?)

oshamash: @lamby yeah, I forgot to redo the comment, talked about it already with @michael-grunder 
either way personally I'm against doing `abort()` if we can handle it, for example when doing callback-allocation, we can return error instead of abort, remember that this is a 3rd-party library for most users, and crashing someone's application due to internal errors is not the way to do things IMHO

PR #756, opened 2020-01-29
install alloc.h

ch1aki:
alloc.h added in https://github.com/redis/hiredis/pull/754
But not added to INSTALL_INCLUDE_PATH


michael-grunder:
Merged, nice catch!

Issue #758, opened 2020-02-04
adapters/libevent.h:155: undefined reference to hi_calloc


tomarus:
Hey,

I get this error since a few days. I'm using the libevent adapter. Probably has to do with the recent alloc changes from #754

` /usr/local/include/hiredis/adapters/libevent.h:155: undefined reference to `hi_calloc(unsigned long, unsigned long)'`

Thanks for looking into this :)

michael-grunder:
Almost certainly, but can you provide a bit more information about how you're building hiredis?

I'm not able to reproduce this.  Do you perhaps need a `make install` of hiredis?

tomarus:
Hi, we basically do a clean clone on a fresh docker image (buildpack-deps:buster) during each build.

>  $ git clone https://github.com/redis/hiredis.git
>  Cloning into 'hiredis'...
>  $ make -C hiredis install && rm -rf hiredis
>  make: Entering directory '/builds/src/nnd/hiredis'
>  cc -std=c99 -pedantic -c -O3 -fPIC   -Wall -W -Wstrict-prototypes -Wwrite-strings -Wno-missing-field-initializers -g -ggdb net.c
>  cc -std=c99 -pedantic -c -O3 -fPIC   -Wall -W -Wstrict-prototypes -Wwrite-strings -Wno-missing-field-initializers -g -ggdb hiredis.c
>  cc -std=c99 -pedantic -c -O3 -fPIC   -Wall -W -Wstrict-prototypes -Wwrite-strings -Wno-missing-field-initializers -g -ggdb sds.c
>  cc -std=c99 -pedantic -c -O3 -fPIC   -Wall -W -Wstrict-prototypes -Wwrite-strings -Wno-missing-field-initializers -g -ggdb async.c
>  cc -std=c99 -pedantic -c -O3 -fPIC   -Wall -W -Wstrict-prototypes -Wwrite-strings -Wno-missing-field-initializers -g -ggdb read.c
>  cc -std=c99 -pedantic -c -O3 -fPIC   -Wall -W -Wstrict-prototypes -Wwrite-strings -Wno-missing-field-initializers -g -ggdb sockcompat.c
>  cc -std=c99 -pedantic -c -O3 -fPIC   -Wall -W -Wstrict-prototypes -Wwrite-strings -Wno-missing-field-initializers -g -ggdb alloc.c
[...]

Issue #750, opened 2020-01-15
extern "C" missing from hiredis_ssl.h


azmtnruz:
I think hiredis_ssl.h header file needs:

```
#ifdef __cplusplus
extern "C" {
#endif
```

and

```
#ifdef __cplusplus
}
#endif
```

So that a c++ app can link to the libhiredis_ssl.so. The hiredis.h header file uses extern "C".

michael-grunder:
We're about to merge a security related patch but once that's merged I'll merge this change into master.

michael-grunder:
Closing via #759

PR #759, opened 2020-02-04
Adds extern C if we detect C++

michael-grunder:
Pretty straightforward but another set of eyes is always good.

See: #758, #750


michael-grunder:
The failing test appears unrelated to the PR.  I triggered a Travis build from the current master and the Windows test fails there as well.

michael-grunder:
Closing in favor of #764

PR #762, opened 2020-02-06
Add CFLAG to be able run the tests against cluster

borislitvinov:


michael-grunder:
I don't totally understand what the purpose of this is.

I see it skips the unix socket and timeout test if `REMOTE_DB` is defined, but how does this pertain to cluster?

borislitvinov:
Exactly . DMC - it is our access point to cluster. 
1. We can;t access the socket ( the test fails constantly)
2. DMC doesn't support commands like "DEBUG"  . So the tests which are using those commands fail as well
In order to skip it i'll compile the executable with REMOTE_DB flag

michael-grunder:
> We can;t access the socket ( the test fails constantly)

It seems reasonable to check for the existence of the unix socket and skip that test if it's not found.

> DMC doesn't support commands like "DEBUG" .

I'm not familiar with `DMC`.  When you say "cluster" do you mean Redis Cluster, or does it mean something different in this context?  The reason I ask is that a whole lot of tests are going to fail if Redis is responding with `MOVED`.

I sketched out a quick [POC](https://github.com/michael-grunder/hiredis/tree/tests-detect-functionality) to do a bit of feature detection in the tests.

What do you think about something like that instead?

borislitvinov:
1. Cluster  - in terms of RS Cluster
> It seems reasonable to check for the existence of the unix socket and skip that test if it's not found.
Agree -> will do it

michael-grunder:
Closing this in favor of [this branch](https://github.com/michael-grunder/hiredis/tree/trying) that has logic to skip unix socket and `DEBUG SLEEP` tests when we don't have them.

Issue #761, opened 2020-02-05
Failing Travis-CI builds in Windows


michael-grunder:
It looks like there is an issue with the Windows build after travis upgraded their workers from `1803` to `1809`.

[Travis-CI community forum post](https://travis-ci.community/t/vcvarsall-bat-freezes-on-new-1809-based-windows-images/7098)

I'm not familiar enough with Windows to be much help, but I was able to get a successful build by building with cmake directly, rather than cmake and ninja.

[Working build in my repo](https://github.com/michael-grunder/hiredis/tree/trying).

michael-grunder:
Closing in favor of #764

PR #764, opened 2020-02-09
Housekeeping fixes

michael-grunder:
This PR contains housekeeping/CI fixes.

* Fixes a currently failing Travis-CI build on Windows.  It's unclear to me what's causing the problem, but at least this way we'll actually have a functional Windows build test. (#761)
* Add `extern "C"` to `alloc.h` and `hiredis_ssl.h` (#750, #759)
* Adds `alloc.h` to the CMakeLists.txt and creates a Windows `.def` file. (#760).


michael-grunder:
I'll squash this when we merge it so the commit history is clean.

PR #765, opened 2020-02-11
install alloc.h when using cmake

fengggli:
Similar to #756, but for cmake build
note: alloc.h added in #754


michael-grunder:
Hi @fengggli, thanks for the PR.

This change is included in #764 so I'm going to close this one in favor of that.

I'm planning on merging that PR today.

PR #768, opened 2020-02-26
Provides an optional cleanup callback for async data.

heronr:
This change adds a cleanup event to the redisAsyncContext such that the private data can be deleted or modified as a result of the context being free'd. This way it is possible to allocate private data on the context whose lifetime is tied to that of the context.

This is essentially the only meaningful change to hiredis made by hiredis-vip to support cluster. Adding the data cleanup event to hiredis proper allows for hiredis-vip to be a pure wrapper class that depends on hiredis directly. This way hiredis-vip won't continually lag behind.

### DIFF
--- a/async.c
+++ b/async.c
@@ -109,6 +109,7 @@ static redisAsyncContext *redisAsyncInitialize(redisContext *c) {
     ac->err = 0;
     ac->errstr = NULL;
     ac->data = NULL;
+    ac->dataCleanup = NULL;
 
     ac->ev.data = NULL;
     ac->ev.addRead = NULL;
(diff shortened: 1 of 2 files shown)


michael-grunder:
I don't have a problem with this.

@mnunberg See any downsides here?

Also you can disregard the failing Windows test, that's not related to your PR.

heronr:
Out of curiosity, @michael-grunder or @mnunberg, are there larger guidelines or suggestions to contributing to hiredis other than blind submitting PRs? I would have liked to have discussed whether this change made sense beforehand but couldn't find the best place to do that.

michael-grunder:
There aren't really any explicit guidelines.  Larger scale changes often come from requirements in Redis (e.g. RESP3 support), but other than that the library is pretty low velocity.

As for other changes, I tend to try and balance the benefits with how much pain it's going to cause when backported into Redis.

That said, you're welcome to open issues to discuss changes.  I'm generally around to respond.

michael-grunder:
Merged, thanks!

PR #771, opened 2020-03-05
change data to be an union, so that it can support uint64 data in 32-bit OS

wond12:
change data to be an union, so that it can support uint64 data in 32-bit OS


michael-grunder:
The `data` member is a `void*` meaning you can put anything you like there, including your union.

Issue #775, opened 2020-03-22
[BUG][MinGW64] Error setting socket timeout 


dellorogiulio:
Hi, 
I'm recently working with the [redis-plus-plus](https://github.com/sewenew/redis-plus-plus) repo which wraps hiredis and I found a very annoying bug: setting a socket timeout of 10**ms**, it becomes 10**s**. I'm running MinGW under Windows.
Analyzing the problem with @sewenew, which is the creator of redis-plus-plus, we could have found the bug.
Let me summarize the problem:
**net.c**
```
316    int redisContextSetTimeout(redisContext *c, const struct timeval tv) {
317        const void *to_ptr = &tv;
318        size_t to_sz = sizeof(tv);
319    #ifdef _WIN32
320        DWORD timeout_msec = tv.tv_sec * 1000 + tv.tv_usec / 1000;
321        to_ptr = &timeout_msec;
322       to_sz = sizeof(timeout_msec);
323    #endif
324        if (setsockopt(c->fd,SOL_SOCKET,SO_RCVTIMEO,to_ptr,to_sz) == -1) {
325            __redisSetErrorFromErrno(c,REDIS_ERR_IO,"setsockopt(SO_RCVTIMEO)");
326            return REDIS_ERR;
327        }
328        if (setsockopt(c->fd,SOL_SOCKET,SO_SNDTIMEO,to_ptr,to_sz) == -1) {
329            __redisSetErrorFromErrno(c,REDIS_ERR_IO,"setsockopt(SO_SNDTIMEO)");
330            return REDIS_ERR;
331        }
332       return REDIS_OK;
333    }
```
Line 320 a `DWORD timeout_msec` is correctly computed from the  `timeval tv ` 
Line 321 `to_ptr` is assign to the address of `timeout_msec` (`to_ptr` is di facto a `DWORD*`)
Line 324 `to_ptr` is pass to ` setsockopt` which is  `win32_setsockoption`

**sockcompat.c**
```
212    int win32_setsockopt(SOCKET sockfd, int level, int optname, const void *optval, socklen_t optlen) {
213        int ret = 0;
214        if ((level == SOL_SOCKET) && ((optname == SO_RCVTIMEO) || (optname == SO_SNDTIMEO))) {
215            struct timeval *tv = optval;
216            DWORD timeout = tv->tv_sec * 1000 + tv->tv_usec / 1000;
217            ret = setsockopt(sockfd, level, optname, (const char*)&timeout, sizeof(DWORD));
[...]

PR #776, opened 2020-03-30
Win32 tests and timeout fix

michael-grunder:
This PR adds three things:

* Working unit tests in Windows
* A Windows timeout fix (see: #775, 58d1654)
* Test detection (see: #762)

When running unit tests in Windows I found two places where functionality differed.

1. In Linux attempting to [connect to a non-existent port](https://github.com/michael-grunder/hiredis/blob/2b652d73ffee6b7cc1c1a9843efb475cf415fefa/test.c#L516) returns an error, but in Windows it just hangs forever.  I believe this is related to a bug in [WSAPoll](https://daniel.haxx.se/blog/2012/10/10/wsapoll-is-broken/) but I really don't know enough about the Windows API to say for sure.  It sure seems related though.

2. Timeouts have slightly different behavior on Windows ([example 1](https://github.com/michael-grunder/hiredis/blob/5679746bf62bca848bd22a715d3304dd6e587489/test.c#L656), [example 2](https://github.com/michael-grunder/hiredis/blob/5679746bf62bca848bd22a715d3304dd6e587489/test.c#L734)).  It seems as if we should honor the same contract in either OS but OTOH people are almost certainly relying on the current behavior.  Interested in opinions.

The test detection logic helps with running the unit tests in Windows (e.g. it skips AF_UNIX tests) but also makes it slightly nicer to run tests if your server isn't local (see #762 as mentioned above).


michael-grunder:
@mnunberg There aren't any changes to the core library here, but let me know if you'd like some modifications w/r/t the Windows changes.

I feel like the correct solution would be to provide the same contract on Windows and Linux in terms of the timeout errors, but I'm sure it will break people's code.

PR #780, opened 2020-04-07
CMake: Set hiredis_ssl shared object version.

yossigo:
This fixes a CMake issue that results with `libhiredis_ssl.so` not having its version set properly (i.e. like `libhiredis.so.0.15`).


michael-grunder:
Merged, thanks!

PR #730, opened 2019-11-17
add static library

animic:

PR #792, opened 2020-04-20
Adds building static libraries

michael-grunder:
Adds building static libraries in both Linux and Windows.

Note: This is the #730 PR but with the Windows `^M` characters stripped.


michael-grunder:
I've tested this in Linux and Windows and it works for me in both environments.

The only real hesitation I have has to do with any possible licence incompatibilities although [it would appear BSD-3 and Apache are compatible](https://softwareengineering.stackexchange.com/questions/40561/is-bsd-license-compatible-with-apache)

michael-grunder:
Going to close this issue and create a new PR if I can implement the functionality without licensing issues

PR #795, opened 2020-04-22
Support timeouts in libev adapater

michael-grunder:
Adds the ability to set a timeout when using the libev adapter.

[The libev docs around timeouts](https://metacpan.org/pod/distribution/EV/libev/ev.pod#Be-smart-about-timeouts).

They describe a variety of ways to handle timeouts with the event lib, and I went with option 2.

### DIFF
--- a/adapters/libev.h
+++ b/adapters/libev.h
@@ -125,7 +125,7 @@ static void redisLibevTimeout(EV_P_ ev_timer *timer, int revents) {
     redisAsyncHandleTimeout(e->context);
 }
 
-static void redisLibevScheduleTimer(void *privdata, struct timeval tv) {
+static void redisLibevSetTimeout(void *privdata, struct timeval tv) {
     redisLibevEvents *e = privdata;
     struct ev_loop *loop = e->loop;
     ((void)loop);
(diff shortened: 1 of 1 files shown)


adapters/libev.h
```diff
@@ -103,13 +104,41 @@ static void redisLibevDelWrite(void *privdata) {
...
     free(e);
 }
 
+static void redisLibevTimeout(EV_P_ ev_timer *timer, int revents) {
+    ((void)revents);
+    redisLibevEvents *e = timer->data;
+    redisAsyncHandleTimeout(e->context);
+}
+
+static void redisLibevScheduleTimer(void *privdata, struct timeval tv) {
```
yossigo, line 128: @michael-grunder please note that currently only `libevent.h` implements a timer and it uses the name `redisLibeventSetTimeout()`, please consider adopting the same naming convention across adapters.

adapters/libev.h
```diff
@@ -136,8 +164,13 @@ static int redisLibevAttach(EV_P_ redisAsyncContext *ac) {
     ac->ev.addWrite = redisLibevAddWrite;
     ac->ev.delWrite = redisLibevDelWrite;
     ac->ev.cleanup = redisLibevCleanup;
+    ac->ev.scheduleTimer = redisLibevScheduleTimer;
     ac->ev.data = e;
 
+    if (ac->c.timeout->tv_sec != 0 || ac->c.timeout->tv_usec != 0) {
+        redisLibevScheduleTimer(e, *ac->c.timeout);
+    }
+
```
yossigo, line 173: As discussed, please consider having this aligned with the libevent behavior (or vice versa if you find the current libevent behavior a bug).

adapters/libev.h
michael-grunder: Updated

adapters/libev.h
michael-grunder: Removed these lines to align with our libevent adapter.

Issue #794, opened 2020-04-21
Hard-coded limit of nested reply depth


jeffreylovitz:
Hello,

Hiredis has the following restriction:
https://github.com/redis/hiredis/blob/0501c623c91344e54cb2775a91509650960789b1/read.c#L418-L423

This is a limitation that [RedisGraph](https://github.com/redisgraph/redisgraph) can hit when processing complex but sane queries.

Could this limit be relaxed?

maguec:
:+1:

michael-grunder:
I think it's reasonable to remove this limit, and certainly before we release v1.0.0.

jeffreylovitz:
@michael-grunder That's fantastic, thanks!

michael-grunder:
I took a crack at removing the depth limitation over on my fork of hiredis.

The change switches from a hard-coded stack to a dynamic one that grows as needed.  I would have expected this to degrade performance (due to the worse cache locality) but it appears offset by the smaller `redisReader` size, because [the benchmarks are nearly identical](https://gist.github.com/michael-grunder/a43301bc70832829ff2f77643cf2617b).

I'd love for other people to give it a try though.  If it seems reasonably stable I'll open a PR here and we can more formally discuss the change.

Cheers,
Mike

jeffreylovitz:
I'd be happy to do some testing! Would you prefer to share scripts like `raw-bench-unlimited-depth` for comparability, or should I test as I see fit?

michael-grunder:
I just put together a quick and dirty repository to directly test the changes to hiredis.

I cleaned it up a bit and [uploaded it here](https://github.com/michael-grunder/reader-bench)

The only place where I'm seeing a noticeable performance degradation is when I literally do something like this:

```c
size_t benchReader(const char *protofile) {
    redisReader *rdr = redisReaderCreate();
    size_t count = 0;
    char buffer[32768];
    size_t read;
    FILE *fp = openOrAbort(protofile);

    while ((read = fread(buffer, 1, sizeof(buffer), fp)) != 0) {
        redisReaderFeed(rdr, buffer,  read);
        count += consumeReplies(rdr);
    }

    if (read > 0) {
        redisReaderFeed(rdr, buffer, read);
        count += consumeReplies(rdr);
    }

    redisReaderFree(rdr);
    fclose(fp);

    return count;
}
```
[...]

Issue #421, opened 2016-04-30
Remove the depth limitation


luin:
When parsing a nested result with depth greater than 7, hiredis will return an error (https://github.com/redis/hiredis/blob/d3c7df33a1e278c6f0ae86506e0cbe5de35e36e3/read.c#L310-L314) which doesn't make sense since it's a common case to deal with a nested array (e.g. tree structure).

I found a related pull request: https://github.com/redis/hiredis/pull/70. Should we just remove this limitation now?

michael-grunder:
Closing in favor of #794

PR #797, opened 2020-04-28
Remove nested depth limitation.

michael-grunder:
This commit removes the nested multi-bulk depth limitation of 7.
We do this by switching to pointer to pointer indirection and
growing the stack in chunks when needed.

Initial benchmarking indicates the worse cache locality seems to be
offset by the reduction in overall redisReader size, resulting in
nearly identical parsing performance to the exiting master branch.

See: #794, #421

### DIFF
--- a/read.c
+++ b/read.c
@@ -46,6 +46,9 @@
 #include "sds.h"
 #include "win32.h"
 
+/* The initial size of our working stack */
+#define REDIS_READER_STACK_SIZE 9
+
 static void __redisReaderSetError(redisReader *r, int type, const char *str) {
     size_t len;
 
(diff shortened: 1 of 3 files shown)


michael-grunder:
For an extensive discussion and lots of benchmarks see #794 

I did quite a big of testing including running the new code using [https://github.com/ralight/mallocfail](mallocfail) as the allocator and everything looks good.

Would love input from @yossigo and/or @mnunberg

yatsukhnenko:
@michael-grunder why do you use `redisReadTask **task;` but not `redisReadTask *task;` ?

read.c
```diff
@@ -46,6 +46,9 @@
 #include "sds.h"
 #include "win32.h"
 
+/* The initial size of our working stack */
```
yossigo, line 49: @michael-grunder Please consider updating the comment, this is not just the initial size but also step size when growing.

michael-grunder: Good catch!  Updated.

PR #800, opened 2020-05-07
Allow users to replace allocator and handle OOM everywhere.

michael-grunder:
This PR replaces every allocation and deallocation with `hi_` variants that are defined in [alloc.h](https://github.com/michael-grunder/hiredis/blob/58e72a42bf7b6ebd132845b1e2d8e2bc56fb03f1/alloc.h#L31)

Additionally, it attempts to gracefully handle OOM everywhere.

To test the new logic, I created "fail allocators" which could be configured to fail after `N` allocations or randomly.

Then I ran our tests under Valgrind while incrementing this counter until I detected a Valgrind error (leak, invalid read/write, etc) or all the tests passed.  We don't have async unit tests so I created some and repeated the process.

A couple points of interest:

*  When fuzz testing hiredis I found a bug where `redisReadTask->elements` was defined as an `int` but `redisReply->elements` was defined as a `size_t`, which could overflow (3e8835e25bfa5ed9dc7fc93cd65d5fdb05a1f80c).
* I fixed a Unix socket [reconnection leak](https://github.com/michael-grunder/hiredis/blob/58e72a42bf7b6ebd132845b1e2d8e2bc56fb03f1/net.c#L561)
* I made a small change to ssl.c to avoid a potential memory leak (8fe3dd78669ec8a3ca3787efe069ac65f99a28eb)

Cheers!


michael-grunder:
One more thing before I forget.

I think we may want to impose a sane limit on multi-bulk elements.  I needed to do that locally or fuzz testing would immediately break hiredis by sending something like:

```*35184372088832\r\n```

To `redisReader`.  This isn't invalid and fits in a 64 bit integer, but it's an absurd count that ends up blocking hiredis for minutes.  It seems unlikely we'd ever need to actually process that many elements.

To fix it locally for fuzz testing I just did this:

```patch
diff --git hiredis.c hiredis.c
index 6ce3049..37795d1 100644
--- hiredis.c
+++ hiredis.c
@@ -44,6 +44,9 @@
 #include "async.h"
 #include "win32.h"

+/* A sane value for maximum number of array elements */
+#define HIREDIS_MAX_ARRAY_ELEMENTS ((1LL<<32) - 1)
+
 static redisContextFuncs redisContextDefaultFuncs = {
     .free_privdata = NULL,
     .async_read = redisAsyncRead,
@@ -150,6 +153,9 @@ static void *createStringObject(const redisReadTask *task, char *str, size_t len
 static void *createArrayObject(const redisReadTask *task, size_t elements) {
     redisReply *r, *parent;

+    if (elements > HIREDIS_MAX_ARRAY_ELEMENTS)
+        return NULL;
+
     r = createReplyObject(task->type);
     if (r == NULL)
         return NULL;
@@ -171,6 +177,7 @@ static void *createArrayObject(const redisReadTask *task, size_t elements) {
                parent->type == REDIS_REPLY_SET);
         parent->element[task->idx] = r;
     }
+
     return r;
 }
```

net.c
```diff
@@ -328,6 +328,22 @@ int redisContextSetTimeout(redisContext *c, const struct timeval tv) {
     return REDIS_OK;
 }
 
+static int _redisContextUpdateTimeout(redisContext *c, const struct timeval *timeout) {
+    /* Same timeval struct, short circuit */
+    if (c->timeout == timeout)
```
yossigo, line 333: @michael-grunder You probably meant to compare the actual value and not the pointer here.

net.c
```diff
@@ -352,20 +368,18 @@ static int _redisContextConnectTcp(redisContext *c, const char *addr, int port,
      * This is a bit ugly, but atleast it works and doesn't leak memory.
      **/
     if (c->tcp.host != addr) {
```
yossigo, line 370: Looks like another case of comparing the pointer rather than the value.

net.c
```diff
@@ -375,10 +389,10 @@ static int _redisContextConnectTcp(redisContext *c, const char *addr, int port,
     }
 
     if (source_addr == NULL) {
-        free(c->tcp.source_addr);
+        hi_free(c->tcp.source_addr);
         c->tcp.source_addr = NULL;
     } else if (c->tcp.source_addr != source_addr) {
```
yossigo, line 394: Same as above.

net.c
```diff
@@ -521,25 +540,30 @@ int redisContextConnectUnix(redisContext *c, const char *path, const struct time
         return REDIS_ERR;
 
     c->connection_type = REDIS_CONN_UNIX;
-    if (c->unix_sock.path != path)
+    if (c->unix_sock.path != path) {
```
yossigo, line 543: Another one.... Looks like a pattern and this one leaks as well.

yossigo:
@michael-grunder Another thing, regarding `HIREDIS_MAX_ARRAY_ELEMENTS ` do you see any reason not to make this runtime configurable rather than a compile time define?

I think it's the kind of things that will need to be changed once in a million, but if that happens it's much easier to just do some runtime configuration rather than full library rebuild.

net.c
michael-grunder: Good catch on the leak, I'll update the logic!

alloc.h
```diff
@@ -31,23 +31,10 @@
...
-char *hi_strdup(const char *str);
-
-#ifdef __cplusplus
-}
-#endif
+#define hi_malloc malloc
+#define hi_calloc calloc
+#define hi_realloc realloc
+#define hi_strdup strdup
+#define hi_free free
```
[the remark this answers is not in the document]
michael-grunder: Sure, that works for me.  It would make it much simpler for end-users to plug in their own functions too.

michael-grunder:
Thanks for going through the code!

> @michael-grunder Another thing, regarding HIREDIS_MAX_ARRAY_ELEMENTS do you see any reason not to make this runtime configurable rather than a compile time define?

I'm happy to add this as configurable similar to our reader `maxbuf`. 

---

The pointer comparisons were intentional but this is what we're doing now.  I think it was initially done this way to handle reconnections.

AFAIK it's safe to treat two pointers equal if they point to the same address and are of the same type.  Do you mind clarifying what your concern is?  It's possible I'm totally missing your point :smile:

For reference, this logic has been in place for years ([example](https://github.com/redis/hiredis/blame/685030652cd98c5414ce554ff5b356dfe8437870/net.c#L294), [example](https://github.com/redis/hiredis/blame/eafb085d1172bb8aa1cdfd230d580910a691ecd4/net.c#L524)).

---

1.  I'll add `HIREDIS_MAX_ARRAY_ELEMENTS` as a configurable runtime setting, with a sane default.
2.  I'll switch to pluggable heap allocation functions that will default to `malloc`, `calloc`, `free`, etc.

yossigo:
@michael-grunder Yes I see the reconnect now! So this obviously works for reconnects, but if a user calls this function with the same value (but not same pointer of course) it will get reallocated. Anyway, I guess that's not related to this PR.

alloc.h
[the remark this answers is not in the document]
michael-grunder: Updated the logic to use a single pass-through struct with function pointers and just access the appropriate member depending on what kind of allocation it is.

I went with static inline functions here because it allowed me to have the allocation functions configurable without paying a performance price.  Using non-inline functions defined in `alloc.c` cost between 8-15% on my machine for some workloads.

michael-grunder:
Hi @yossigo, would you like to see any changes in this PR?

alloc.c
```diff
@@ -0,0 +1,95 @@
...
+
+hiredisAllocFuncs hiredisAllocFns = {
+    .malloc = malloc,
+    .calloc = calloc,
+    .realloc = realloc,
+    .strdup = strdup,
+    .free = free,
+};
+
+/* Override one or more of hireids' allocators */
```
yossigo, line 44: @michael-grunder Maybe I'm a bit over cautious here (Paranoid?), but this leaves some room for hard to find issues in case someone only substitutes some of these functions. WDUT?

I'd feel somewhat safer if we do all-or-nothing, so if a used function is left out we'll just hit a null pointer exception and it'll be obvious what went wrong.

michael-grunder: I think you're right.  It's bound to cause confusion.

Issue #809, opened 2020-05-17
Makefile does not install TLS libraries 


theDogOfPavlov:
### What do I see?

When I run the make and install, the tests complete as expected, however the TLS headers and libraries are not copied over.

### What do I expect to see?

The headers and libraries should be installed as expected.

### To reproduce
build
```
make USE_SSL=1
make install
```
TLS headers and libraries created as expected, but the Makefile does not copy them to the right locations
```
mkdir -p /usr/local/include/hiredis /usr/local/include/hiredis/adapters /usr/local/lib
cp -pPR hiredis.h async.h read.h sds.h alloc.h /usr/local/include/hiredis
cp -pPR adapters/*.h /usr/local/include/hiredis/adapters
cp -pPR libhiredis.so /usr/local/lib/libhiredis.so.0.15
cd /usr/local/lib && ln -sf libhiredis.so.0.15 libhiredis.so
cp -pPR libhiredis.a /usr/local/lib
mkdir -p /usr/local/lib/pkgconfig
cp -pPR hiredis.pc /usr/local/lib/pkgconfig
```

michael-grunder:
I see the same thing.  I'll take a closer look later today after I've had coffee :smile:

michael-grunder:
I got this working over on my fork.  I'm not sure this is the cleanest solution but it seems like both:

```bash
USE_SSL=1 make && sudo make install
```

and

```bash
sudo USE_SSL=1 make install
```

Should work to build and install the SSL libraries.

PR #812, opened 2020-05-19
Fix missing SSL build/install options.

yossigo:
* Use different build commands for non-Linux linkers.
* Add missing install targets.


michael-grunder:
I was messing around with this to fix #809 [over on my fork](https://github.com/michael-grunder/hiredis/tree/install-ssl)

Are we sure we don't need to update the `soname` to hiredis_ssl.so for the SSL shared object?  I didn't at first but when I `sudo USE_SSL=1 make install` everything installs correctly, but then `ldconfig` spits out a warning and `ld` can't find `redisSecureConnection` (at least for me).

```bash
$ sudo make USE_SSL=1 install
# < hiredis.so + hiredis_ssl.so are installed >
$ sudo ldconfig                       
/sbin/ldconfig.real: /usr/local/lib/libhiredis.so.0.15 is not a symbolic link

$ cc onlyssl.c -lhiredis -lhiredis_ssl
/usr/bin/ld: /tmp/ccBObdNY.o: in function `main':
onlyssl.c:(.text+0xd8): undefined reference to `redisSecureConnection'
collect2: error: ld returned 1 exit status
```

For me, using a distinct [hiredis_ssl.so](https://github.com/michael-grunder/hiredis/blob/install-ssl/Makefile#L62) for the soname made everything work.

Also, someone should give you write access to this repo.  I'd do it but I don't have admin rights.

PR #814, opened 2020-05-20
Use standrad isxdigit instead of custom helper function.

tryfinally:
Standard function available and on most platforms uses a lookup table and not
a check with 3 ranges.


michael-grunder:
Merged, thanks!

PR #823, opened 2020-05-25
Use unique names for allocator struct members

michael-grunder:
Using `strdup` as a struct member causes issues in older gcc

cc @carlosabalde I think this should fix it for you (it fixed it for me with gcc-5.4.0 in a xenial docker container)

[See context here](https://github.com/redis/hiredis/commit/8e0264cfd6889b73c241b60736fe96ba1322ee6e#commitcomment-39413269)


carlosabalde:
Yes, fixed. Thanks!

PR #826, opened 2020-06-01
Use explicit pointer casting for c++ compatibility

aureus1:
Small fix to be able to build with g++ without -fpermissive


michael-grunder:
Merged, thank you!

Issue #722, opened 2019-10-18
BRPOP... 0 (and similar) now timeout after connection timeout rather than block


shuckc:
I have a c program that does a blocking pop from a redis list to retrieve work. In order to prevent long connection timeouts, I set a connection timeout.

    redisContext *rc = NULL;
    struct timeval redis_timeout = { 1, 500000 }; // 1.5 seconds
    redisReply *reply;
    ... 
    rc = redisConnectWithTimeout(optarg, 6379, redis_timeout);
    if (rc == NULL || rc->err) {
      if (rc) {
        printf("Redis connection error: %s\n", rc->errstr);
        redisFree(rc);
      } else {
        printf("Redis connection error: can't allocate redis context\n");
      }
      exit(1);
    }
    ...
    reply = redisCommand(rc, "BRPOP tokens 0");
    if (rc->err) {
      printf("Connection error on wakeup: %s\n", rc->errstr);
      exit(1);
    }
    freeReplyObject(reply);
 
Using hiredis `master`, the `Connection error on wakeup` fires after 1.5s.
If I wanted an _operation_ timeout, i would use a non-zero final argument to BRPOP

If I build this against v14, it blocks until the server has an item pushed into the list.

    hiredis:
        git clone https://github.com/redis/hiredis.git hiredis
        cd hiredis && git -c advice.detachedHead=false checkout v0.14.0

From a quick review of master I think this is a regression with the new SSL support, although I will need to bisect it to be sure.

shuckc:
Reproducing code:

$ cat Makefile 
```
LIBS=-lm
CFLAGS=-ggdb -O3 -Wall -I./hiredis
OBJS=blocker.o

blocker: blocker.o hiredis/libhiredis.a
	$(CC) $^ -o $@ $(CFLAGS) $(LIBS)

clean:
	rm -f $(OBJS)
	rm -Rf hiredis

blocker.o: hiredis blocker.c

hiredis:
	git clone https://github.com/redis/hiredis.git hiredis
	cd hiredis && git -c advice.detachedHead=false checkout master
	make -C hiredis

```

$ cat blocker.c 
```
#include <stdio.h>
#include <stdlib.h>
#include <hiredis.h>

int main(int argc, char **argv) {
  redisContext *rc = NULL;
  struct timeval redis_timeout = { 1, 500000 }; // 1.5 seconds
  redisReply *reply;

  rc = redisConnectWithTimeout("localhost", 6379, redis_timeout);
[...]

PR #829, opened 2020-06-02
timeout option in redisConnectWithOptions should be on the connect op…

valentinogeron:
…eration only

in order to set timeout for commands we should use redisSetTimeout

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -790,9 +790,7 @@ redisContext *redisConnectWithOptions(const redisOptions *options) {
         // Unknown type - FIXME - FREE
         return NULL;
     }
-    if (options->timeout != NULL && (c->flags & REDIS_BLOCK) && c->fd != REDIS_INVALID_FD) {
-        redisContextSetTimeout(c, *options->timeout);
-    }
+
     return c;
 }
 


valentinogeron:
when a timeout is specified as an option for `redisConnectWithOptions` (or its equivalents e.g `redisConnectWithTimeout`), it configures the created `redisContext` to use this timeout for command as well.
The current implementation is not backward compatible and may break applications.

My suggestion is that if the application would choose to use a timeout for commands as well, it will have to call `redisSetTimeout` explicitly.

@michael-grunder what do you think?

michael-grunder:
I tend to agree they should be distinct (Related issue: #722).

We are able to break things (since we've not yet released v1.0.0) but once we do we'll be stuck with it until 2.0.0.

I lean toward merging the PR but I suppose we could to a third route and add a distinct `command_timeout` to `redisContextOptions` so the user could specify both.

@yossigo Any preference?

yossigo:
@michael-grunder I agree, I think this PR is a bugfix because there's really no justification to break compatibility. An explicit command timeout seems like a good idea regardless of the fix.

yossigo:
@michael-grunder just to confirm with you, I think this fix should be part of 1.0.0 and the command_timeout improvement is nice to have but not must. What do you think?

michael-grunder:
Yes, I'll make sure to merge this before we release v1.0.0.

I want to implement proper handling for `>` PUSH messages as well but then I think we can call v1.0.0 feature complete.

michael-grunder:
Merged, thanks @valentinogeron!

PR #836, opened 2020-06-21
fix #785: defer TCP_NODELAY in async tcp connections

### LINKED ISSUE #785: setsockopt(TCP_NODELAY) in Windows
There appears to be a Windows-specific issue setting TCP_NODELAY in non-blocking mode.

See related issues #736, #779.

Yielding the timeslice (with `_sleep(0)`) solves the problem, but I have not been able to verify if it is the *correct* way to solve the problem.

If it is correct, perhaps adding it to win32_setsockopt is the cleanest solution:

```c
diff --git sockcompat.c sockcompat.c
index f99d14b..b7990b8 100644
--- sockcompat.c
+++ sockcompat.c
@@ -216,6 +216,10 @@ int win32_setsockopt(SOCKET sockfd, int level, int optname, const void *optval,
         DWORD timeout = tv->tv_sec * 1000 + tv->tv_usec / 1000;
         ret = setsockopt(sockfd, level, optname, (const char*)&timeout, sizeof(DWORD));
     } else {
+#if _MSC_VER
+        /* Windows workaround */
+        if (optname == TCP_NODELAY) _sleep(0);
+#endif
         ret = setsockopt(sockfd, level, optname, (const char*)optval, optlen);
     }
     _updateErrno(ret != SOCKET_ERROR);
```

@mbitsnbites Do you mind weighing in as you wrote the windows compatibility layer?

Also curious if @mnunberg or @yossigo have opinions one way or the other.

OmriSteiner:


michael-grunder:
Hi @amitv87, @chris-shenzhen-china, @hunterfortuin,

It would be great if you could test this PR in Windows and confirm it solves the issue.  I'd love to get this fixed before we release v1.0.0.

cc @yossigo I'll give the PR a good test but it looks likely this solves the issue.

OmriSteiner:
I've sanity-tested this on Windows 10 & Ubuntu 20.04.

michael-grunder:
Awesome, thank you!  

It's just that I only have Windows in a VM so it's good to get verification from multiple people.

yossigo:
@OmriSteiner @michael-grunder Looks good indeed!

hunterfortuin:
This looks good to me as well!

michael-grunder:
Merged, thank's again!

PR #839, opened 2020-06-29
add a command_timeout to redisOptions

valentinogeron:
the user can now specify both connect_timeout and command_timeout
when creating redis context.

### DIFF
--- a/async.c
+++ b/async.c
@@ -810,19 +810,19 @@ int redisAsyncFormattedCommand(redisAsyncContext *ac, redisCallbackFn *fn, void 
 }
 
 int redisAsyncSetTimeout(redisAsyncContext *ac, struct timeval tv) {
-    if (!ac->c.timeout) {
-        ac->c.timeout = hi_calloc(1, sizeof(tv));
-        if (ac->c.timeout == NULL) {
+    if (!ac->c.command_timeout) {
+        ac->c.command_timeout = hi_calloc(1, sizeof(tv));
+        if (ac->c.command_timeout == NULL) {
             __redisSetError(&ac->c, REDIS_ERR_OOM, "Out of memory");
             __redisAsyncCopyError(ac);
             return REDIS_ERR;
         }
     }
 
-    if (tv.tv_sec != ac->c.timeout->tv_sec ||
-        tv.tv_usec != ac->c.timeout->tv_usec)
+    if (tv.tv_sec != ac->c.command_timeout->tv_sec ||
+        tv.tv_usec != ac->c.command_timeout->tv_usec)
     {
-        *ac->c.timeout = tv;
+        *ac->c.command_timeout = tv;
     }
 
     return REDIS_OK;
--- a/async_private.h
+++ b/async_private.h
@@ -54,15 +54,14 @@
     } while(0);
 
 static inline void refreshTimeout(redisAsyncContext *ctx) {
-    if (ctx->c.timeout && ctx->ev.scheduleTimer &&
-        (ctx->c.timeout->tv_sec || ctx->c.timeout->tv_usec)) {
-        ctx->ev.scheduleTimer(ctx->ev.data, *ctx->c.timeout);
-    // } else {
-    //     printf("Not scheduling timer.. (tmo=%p)\n", ctx->c.timeout);
-    //     if (ctx->c.timeout){
-    //         printf("tv_sec: %u. tv_usec: %u\n", ctx->c.timeout->tv_sec,
-    //                ctx->c.timeout->tv_usec);
-    //     }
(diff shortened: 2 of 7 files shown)


valentinogeron:
@michael-grunder - as we talked in https://github.com/redis/hiredis/pull/829, I added an option for `connect_timeout` and `command_timeout` in `redisOptions` so the user can specify both

michael-grunder:
This is a nice quality of life improvement.

What do you think about a minor [refactor on my branch](https://github.com/michael-grunder/hiredis/commit/23b3f83724eb2a0a9911b906f6d61affa521f35a)?

It's a bit easier for me to follow but doesn't change any functionality.

valentinogeron:
looks good!
small question: what `EL` stands for?

michael-grunder:
I'm guessing it stands for "event library" but I didn't initially name them so I could be wrong.

valentinogeron:
make sense. BTW - there are any other comments?

michael-grunder:
No, it looks good but I'll run it through an async testing setup I've got before merging.

I think we should include it in v1.0.0

valentinogeron:
@michael-grunder - any updates?

michael-grunder:
I'm putting together v1.0.0 and will try to get it included.

I may tweak the change a little bit to use the existing `redisContext->timeout` name, with an additional member `redisContext->command_timeout`.  It's less descriptive but it would make the changes completely backward-compatible (I think.  I need to go through it in more detail to be sure).

michael-grunder:
@yossigo Any objections to merging this for v1.0.0?  Depending on how much of a stickler we want to be to semver I think even an additional struct field is _technically_ a breaking change and would need to wait for 2.0.0

The alternative is to reserve the additional timeout fields in v1.0.0 and push the feature in 1.0.1 which we would able to do while maintaining 100% ABI compatibility.

[I'd merge this version of the PR on my fork](https://github.com/michael-grunder/hiredis/commit/23b3f83724eb2a0a9911b906f6d61affa521f35a).

yossigo:
@michael-grunder I think it can be part of 1.0.0.

BTW I am in favor of being strict with such changes, especially as hiredis traditionally offered basically no encapsulation and all major structs are all over the header files so every little change is a potential ABI breakage.

michael-grunder:
> BTW I am in favor of being strict with such changes, especially as hiredis traditionally offered basically no encapsulation and all major structs are all over the header files so every little change is a potential ABI breakage.

Agreed.  Might be worth looking at using opaque structures for the next major version although that's a long ways off.

michael-grunder:
@valentinogeron Thanks, merged!  I'm going to get this merged into 1.0.0-rc1 and update the docs/changelog.

wbangna:
This change showed one negative effect for me.
I implemented a client that may be configured to use TLS.
When this client tries to connect to a redis port that does not use TLS, the timeout setting does work for the TCP connect but not for the SSL connect. This leads to a blocking client.

    redisSSLContextError ssl_error = 0;
    redis_ssl_ctx = redisCreateSSLContext(
                            ca_filename, nullptr /*capath*/, cert_filename,
                            key_filename, nullptr /*sni*/, &ssl_error);
    
    if (!redis_ssl_ctx || ssl_error != 0)
    {
        /* error handling here */
    }
    
    redis_ctx = redisConnectWithTimeout(server_name, server_port, timeout);
    
    /* Workaround: */
    /* redisContextSetTimeout(redis_ctx, timeout); */
    
    if ( redis_ctx->err != 0 )
    {
        /* error handling here */
    }
    
    if (redisInitiateSSLWithContext(redis_ctx, redis_ssl_ctx) != REDIS_OK)
    {
        /* error handling here */
    }

Is the workaround (commented out) the right solution or should I open an issue?

valentinogeron:
@wbangna  - The PR goal is to have the ability to control the connect timeout and the command timeout separately.
The TLS handshake occurs after a TCP connection is established (the connect timeout is not relevant from this part), so you must specify `command_timeout` as well.

Your workaround will work, but you need to use `redisSetTimeout` instead.
Another option is to use `redisConnectWithOptions`. With this API, you can configure whatever you want. 
For example:
```
struct timeval connect_timeout = { 1, 500000 };
struct timeval command_timeout = { 2, 500000 };

redisOptions options = {0};
REDIS_OPTIONS_SET_TCP(&options, ip, port);
options.connect_timeout = &connect_timeout;
options.command_timeout = &command_timeout;
redisContext *ctx = redisConnectWithOptions(&options);
```

wbangna:
@valentinogeron Thank you for your response. My test was successful with the `redisConnectWithOptions()` function.

PR #840, opened 2020-07-02
pass MSG_NOSIGNAL to send in order to prevent signal generation.

nerdmaennchen:
When the connection to the redis server breaks a call to send will generate a SIGPIPE signal.
In applications where multiple connections are handles this behavior makes handling broken connections really hard.
However, passing MSG_NOSIGNAL to send solves this as proper error propagation is already implemented in the error handling code just after send


nerdmaennchen:
mh, apparently this isn't portable :(

Issue #825, opened 2020-05-29
RESP3 PUSH messages incorrectly use pending callback


lukepalmer:
I think that RESP3 PUSH data can be incorrectly sent to a callback that is waiting for a reply.

async.c makes the following assumption:
```
        /* Even if the context is subscribed, pending regular callbacks will                                                                                                       
         * get a reply before pub/sub messages arrive. */                                                                                                                          
        if (__redisShiftCallback(&ac->replies,&cb) != REDIS_OK) {                                                                                                                  

```
I think this is not true because the RESP3 spec says:

Push data may be interleaved with any protocol data, but always at the top level, so the client will never find push data in the middle of a Map reply for instance.

Clients should normally react to the publication of a push data by invoking a callback associated with the push data type.

michael-grunder:
Yes, at present hiredis can parse `>` push messages but there's no logic to handle those kinds of "out of band" messages properly.

This would help redis-cli too:
antirez/redis/issues/7125

@yossigo I kind of think we should support this before releasing v1.0.0.  I don't think it will be all that involved.

michael-grunder:
I put together a draft for the proper handling of RESP3 `PUSH` messages.

Basically the commit allows users to specify a custom handler for `PUSH` messages if they wish in sync or async hiredis.

In order to maintain backward compatibility, the async API always handles non-subscribe `PUSH` messages out of band but if no callback is supplied by the user all we do is call `freeReplyObject` and continue the `redisGetReply` loop.

I'll do some more robust testing but the branch seems to work pretty well.

michael-grunder:
Closed via #841

PR #841, opened 2020-07-06
Resp3 oob push support

michael-grunder:
Implement proper RESP3 push message support in both sync and async hiredis.

The change intercepts `RESP3` PUSH messages by default in sync hiredis and calls `freeReplyObject` when they are received.  If the user wishes to do something specific with the replies they can either set a handler up in `redisOptions` or specify a handler after connected with `redisSetPushHandler`

This works similarly in async hiredis except we always call `freeObject` regardless of whether the user has specified an additional handler (because async hiredis always frees replies).

It also adds some tests (in Linux) for the handler by using Chris Lea's Redis PPA (so we can test against Redis 6.0.x).

See: #825

cc: @yossigo, @lukepalmer

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -688,7 +688,13 @@ static redisContext *redisContextInit(const redisOptions *options) {
         return NULL;
 
     c->funcs = &redisContextDefaultFuncs;
-    c->push_cb = options->push_cb;
+
+    /* Set any user supplied RESP3 PUSH handler or use freeReplyObject
+     * as a default unless specifically flagged that we don't want one. */
+    if (c->push_cb != NULL)
+        redisSetPushHandler(c, options->push_cb);
+    else if (!(options->options & REDIS_OPT_NO_PUSH_HANDLER))
+        redisSetPushHandler(c, freeReplyObject);
 
     c->obuf = sdsempty();
     c->reader = redisReaderCreate();
--- a/hiredis.c
+++ b/hiredis.c
@@ -184,7 +184,8 @@ static void *createArrayObject(const redisReadTask *task, size_t elements) {
         parent = task->parent->obj;
         assert(parent->type == REDIS_REPLY_ARRAY ||
                parent->type == REDIS_REPLY_MAP ||
-               parent->type == REDIS_REPLY_SET);
+               parent->type == REDIS_REPLY_SET ||
+               parent->type == REDIS_REPLY_PUSH);
         parent->element[task->idx] = r;
     }
     return r;
@@ -687,6 +688,8 @@ static redisContext *redisContextInit(const redisOptions *options) {
         return NULL;
 
     c->funcs = &redisContextDefaultFuncs;
+    c->push_cb = options->push_cb;
+
     c->obuf = sdsempty();
     c->reader = redisReaderCreate();
     c->fd = REDIS_INVALID_FD;
@@ -773,7 +776,7 @@ redisContext *redisConnectWithOptions(const redisOptions *options) {
         c->flags |= REDIS_REUSEADDR;
     }
     if (options->options & REDIS_OPT_NOAUTOFREE) {
-      c->flags |= REDIS_NO_AUTO_FREE;
+        c->flags |= REDIS_NO_AUTO_FREE;
     }
 
     if (options->type == REDIS_CONN_TCP) {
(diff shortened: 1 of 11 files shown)


michael-grunder:
The failing Windows test seems choco related.  I'll see if I can get it working again by switching to memurai

test.c
```diff
@@ -77,6 +79,51 @@ static long long usec(void) {
...
+    sds val;
+    char *s, *e;
+
+    reply = redisCommand(c, "INFO");
+    if (reply == NULL || c->err) {
+        fprintf(stderr, "Error:  Can't execute INFO, aborting!\n");
+        exit(1);
+    }
+
+    if ((s = strstr(reply->str, field)) == NULL || *(s+len) != ':' ||
```
yossigo, line 94: Consider being defensive here and attaching a warning this is not really a generic and safe implementation of `get_info_value()` (as it won't deal with overlapping field names, etc.), and it's good only for the specific test purposes it was created for.

hiredis.c
```diff
@@ -687,6 +688,14 @@ static redisContext *redisContextInit(const redisOptions *options) {
         return NULL;
 
     c->funcs = &redisContextDefaultFuncs;
+
+    /* Set any user supplied RESP3 PUSH handler or use freeReplyObject
```
yossigo, line 692: Doesn't this result with double free with the async context?

test.c
michael-grunder: Sure.  Might as well switch back to the non-generic version until we actually need another `INFO` field.

hiredis.c
michael-grunder: No in [redisAsyncConnectWithOptions](https://github.com/redis/hiredis/pull/841/files#diff-8bd6782b369f6afec2c090b37d553d46R172) I flag us as not wanting a handler unless the user supplies one specifically.

There's still a danger of the user explicitly setting the handler to `freeReplyObject` but I'm not sure how we can avoid that.

yossigo: But `redisProcessCallbacks` calls `__redisRunPushCallback` (which calls `c->push_cb`, which ends up calling `freeReplyObject`), and then calls `c->reader->fn->freeObject` once again.

If the contract with the push callback is consistent with other callbacks, the callback gets a borrowed reply and must not free it. But then the default callback should be `NULL` and do nothing rather than `freeReplyObject`. Am I missing something here?

michael-grunder: There is every possibility I'm missing something :smile:

If you compile and run this demonstration program you'll find `push_cb` is NULL.

```c
#include "hiredis/hiredis.h"
#include "hiredis/async.h"

int main(void) {
    redisAsyncContext *ac;
    ac = redisAsyncConnect("localhost", 6379);

    /* ac->c.push_cb will be NULL */
    printf("Push callback: %p\n", ac->c.push_cb);
}
```

```bash
$ cc -o push-example async-push.c hiredis/libhiredis.a && ./push-example
Push callback: (nil)
```

It only defaults to `freeReplyObject` in synchronous hiredis but only so we don't mess up the order of replies that the caller sees with regular `redisCommand` usage.

If you think it better for the `push_cb` to be something like `void (pushHandler*)(const void *)` that's totally fine with me.  

The reason I don't automatically free the reply in the sync API is that it lets the user make the choice.  If they have a use case where they want to keep it around they can do so without having to copy memory.

yossigo: @michael-grunder you are right of course, my bad!

michael-grunder:
@yossigo If you're good with the API I will get it merged and add some documentation.

Then I can put together v1.0.0 RC1 :tada:

yossigo:
@michael-grunder yes `privdata` could be useful! and RC1 would be great!

michael-grunder:
@yossigo If you have some time I'd love a second pair of eyes on the (famous last words) _final_ reworking of the API.

I've bikeshed this feature to death but don't mind making changes since we're stuck with it once we officially release v1.0.0

TL;DR of the new API

I reworked the callback prototypes:

```c
typedef void (redisPushFn)(void *, void *);
typedef void (redisAsyncPushFn)(struct redisAsyncContext *, void *, void *);
```

As you can see I added `c->privdata` to the arguments I send to the sync callback, and changed the async callback to match our existing `redisCallbackFn` prototype.  I think this will be convenient for users who are using a "generic" callback handler, as they'll be able to use their existing logic with an added check for the `REDIS_REPLY_PUSH` type.

If you think this would be confusing for users it's easy to drop the final argument, [which is currently unused](https://github.com/redis/hiredis/pull/841/files#diff-8bd6782b369f6afec2c090b37d553d46R295).

PR #848, opened 2020-07-17
fix windows compiling with mingw

rmalizia44:


michael-grunder:
This seems safe enough but I wonder why our Travis [MinGW builds](https://travis-ci.org/github/redis/hiredis/jobs/707009194) work just fine.

PR #852, opened 2020-07-22
All warnings as errors

masariello:


michael-grunder:
I'm inclined to close this PR as I don't know if adding `-Werror` is prudent.

If we add the flag, it will mean the library can spontaneously break when new compilers are released, which would be a big pain point for our users.  Also, anyone who wishes to add `-Werror` can do so via `CFLAGS`.

If a new version of `gcc` or `clang` is released that throws errors we should then fix them in the repo as they occur.

PR #855, opened 2020-07-29
Move SSL management to a distinct private pointer.

michael-grunder:
We need to allow our users to use `redisContext->privdata` as context
for any RESP3 PUSH messages, which means we can't use it for managing
SSL connections.

Bullet points:

* Create a secondary redisContext member for internal use only called
  `privctx` and rename the `redisContextFuncs->free_privdata` accordingly.

* Adds a `free_privdata` function pointer so the user can tie allocated
  memory to the lifetime of a redisContext (like they can already do
  with `redisAsyncContext`)

* Enables SSL tests in .travis.yml (On Linux/OSX 64 bit)


michael-grunder:
@yossigo I'm not overly fluent in CMake so if there is a better way to structure the changes feel free to let me know.

PR #864, opened 2020-08-12
TLS: Validate the hostname of the server certificate

mkauf:
Avoid MITM attacks by checking whether the server's certificate contains the server's hostname.

Currently this works only with OpenSSL 1.1.0 or newer.

### DIFF
--- a/hiredis_ssl.h
+++ b/hiredis_ssl.h
@@ -88,8 +88,9 @@ int redisInitOpenSSL(void);
  * certificate and private key files to use for authentication. They need to
  * be both specified or omitted.
  *
- * server_name is an optional and will be used as a server name indication
- * (SNI) TLS extension.
+ * server_name is optional and will be used as a server name indication (SNI)
+ * TLS extension and to validate the hostname of the server's certificate
+ * (this requires OpenSSL 1.1.0 or newer).
  *
  * If error is non-null, it will be populated in case the context creation fails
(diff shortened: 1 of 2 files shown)


michael-grunder:
Hi, thanks for the PR!

The code looks fine to me.  The only question I have is whether users are going to want to make the check optional (as opposed to always happening with OpenSSL >= 1.1.0 when they specify `server_name`).

Is the problem that `SSL_set_tlsext_host_name` doesn't by itself actually do any hostname validation?

@yossigo Any thoughts?

mkauf:
> Is the problem that ``SSL_set_tlsext_host_name`` doesn't by itself actually do any hostname validation?

Yes, it only sets the SNI hostname. This name is not used to validate the server certificate.

yossigo:
@mkauf Thanks for this PR for raising this issue in the first place!

Technically this is a breaking change because anyone who's using SNI but don't have compatible certs would experience connection errors. As hiredis 1.0.0 was just released I think we *may* be okay with that, but would like to raise two points:

1. I think we need an option to disable name checks, even if SNI is used. I'd consider something like `redisSSLContextSetOptions(ctx, REDIS_SSL_NO_HOSTNAME_CHECK)`;
2. OpenSSL 1.0.x is still widely used and we'd want hiredis to behave the same way on both versions.

@michael-grunder Does this make sense to you?

PR #872, opened 2020-09-05
CMakeLists.txt: allow building without a C++ compiler

ffontaine:
Define hiredis as a C project (and use a single `PROJECT` statement) to
avoid the following build failure if a C++ compiler is not found:

```
CMake Error at CMakeLists.txt:3 (PROJECT):
  The CMAKE_CXX_COMPILER:

    /srv/storage/autobuild/run/instance-1/output-1/host/bin/arm-linux-g++

  is not a full path to an existing compiler tool.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```

The only cpp source file is examples/example-qt.cpp which is never
compiled with cmake buildsystem. This file is compiled only with the
Makefile buildsystem so perhaps it should be removed. If it is added to
the cmake buildsystem, a call to `enable_language(CXX)` will have to be
added.

Fixes:
 - http://autobuild.buildroot.org/results/830ec3398cd29b9fc5cde06a225ef531d7a9d850

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,4 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 3.4.0)
-INCLUDE(GNUInstallDirs)
-PROJECT(hiredis)
 
 OPTION(ENABLE_SSL "Build hiredis_ssl for SSL support" OFF)
 OPTION(DISABLE_TESTS "If tests should be compiled or not" OFF)
@@ -20,7 +18,8 @@ getVersionBit(HIREDIS_SONAME)
 SET(VERSION "${HIREDIS_MAJOR}.${HIREDIS_MINOR}.${HIREDIS_PATCH}")
 MESSAGE("Detected version: ${VERSION}")
 
-PROJECT(hiredis VERSION "${VERSION}")
+PROJECT(hiredis LANGUAGES "C" VERSION "${VERSION}")
+INCLUDE(GNUInstallDirs)
 
 # Hiredis requires C99
 SET(CMAKE_C_STANDARD 99)


ffontaine:
Do you have some comments on this PR?

ffontaine:
No feedback on this PR opened more than one year ago?

CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 3.4.0)
 INCLUDE(GNUInstallDirs)
-PROJECT(hiredis)
```
bjosv, line 3: Instead of removing this line I would like to have a
`PROJECT(hiredis LANGUAGES "C")`
before 
`INCLUDE(GNUInstallDirs)`
since `GNUInstallDirs` needs a language to be set:

```
CMake Warning (dev) at /snap/cmake/580/share/cmake-3.17/Modules/GNUInstallDirs.cmake:225 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
Call Stack (most recent call first):
  CMakeLists.txt:2 (INCLUDE)
```

bjosv (commented):
I'm not a maintainer but this seems fine with me in general, just one comment added.
~~We have one example `hiredis-example-qt` that requires C++, but this is not (currently) included in the CMake build.
Maybe this can be handled with CMakes [enable_language](https://cmake.org/cmake/help/latest/command/enable_language.html) when the time comes.~~
Sorry, this was already mentioned, maybe I should start to read everything :)

CMakeLists.txt
ffontaine: Thanks for your review, I updated the PR to move `INCLUDE(GNUInstallDirs)` after the `PROJECT` statement.

bjosv: Ah yes, even better! The provided `CMAKE_INSTALL_` variables weren't needed until the install section a bit later..

bjosv (approved):

michael-grunder:
Thank you for the review @bjosv

I'll get this merged over the weekend.  The issue with cmake fixes is that I'm not very familiar with it.

PR #874, opened 2020-09-09
Enable position-independent code, and add PDB files to packages for MSVC builds

masariello:
Static libraries need this


masariello:
@michael-grunder could this be merged to master please?
Thank you

michael-grunder:
This looks ok to me but I'm less familiar with Windows.  I'll try to get it merged this weekend

michael-grunder:
Merged, thanks.

I did notice that we now have a divergence between make and CMake when building static libraries.  Make will build `libhiredis.a` whereas CMake builds `libhiredis_static.a`.  It seems prudent to use the same name unless there is a reason not to.

masariello:
I don't think that can be done in cmake unfortunately. To build static and dynamic together one has to have different target names and therefore different base filenames. If it can be done then it likely requires quite a bit of twists and turns.

PR #879, opened 2020-09-23
Copy error to redisAsyncContext on timeout

shiyuge:
On async timeout, hiredis sets error to redis context, but not redis async context. This requires users to make an extra indirection to the redisContext struct, which is very confusing.


michael-grunder:
Merged, thanks!

PR #881, opened 2020-09-26
docs: Fix simple typo, termined -> terminated

timgates42:
There is a small typo in sds.c.

Should read `terminated` rather than `termined`.


michael-grunder:
Merged, thanks!

PR #887, opened 2020-10-17
Various fixes and updates to the reply parsing implementation

alexsmith1612:
Summary of the salient changes:
- Fix the unset `len` field when creating RESP3 double objects
- Fix RESP3 double infinity parsing
- Add additional validations when parsing various reply types
- Fix the parent type assertions in certain default reply object creation callbacks
- Add various additional reader test cases
- Implement RESP3 BIGNUM support (it was listed in the README, but not actually implemented besides `#define REDIS_REPLY_BIGNUM`)
- Refactor `seekNewline()` to use `memchr()` (an existing comment described why `strchr()` could not be used, but `memchr()` has no such restrictions)

This can be split into multiple separate pull requests if need be, and some of the constituent commits can be squashed together if desired.

#### ADDITIONAL NOTES

The RESP3 attribute type is similarly described in the README but not actually implemented. I did not attempt an implementation yet due to its differing semantics from the other aggregate types that are *actually* part of relevant reply data. It may make sense to separate attributes from the other aggregate types in `redisReply`, and to continue parsing the next redis reply if a top-level attribute message is parsed.

There is also a currently-unimplemented RESP3 bulk error type, which should also be simple to implement, but there is a decision to be made as to whether or not it should be exposed as a distinct type from `REDIS_REPLY_ERROR` (similar to the distinction between `REDIS_REPLY_STATUS` and `REDIS_REPLY_STRING`).

These discussion are probably worth having in a separate issue.

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -235,6 +235,7 @@ static void *createDoubleObject(const redisReadTask *task, double value, char *s
      * decimal string conversion artifacts. */
     memcpy(r->str, str, len);
     r->str[len] = '\0';
+    r->len = len;
 
     if (task->parent) {
         parent = task->parent->obj;
--- a/hiredis.c
+++ b/hiredis.c
@@ -243,7 +243,8 @@ static void *createDoubleObject(const redisReadTask *task, double value, char *s
         parent = task->parent->obj;
         assert(parent->type == REDIS_REPLY_ARRAY ||
                parent->type == REDIS_REPLY_MAP ||
-               parent->type == REDIS_REPLY_SET);
+               parent->type == REDIS_REPLY_SET ||
+               parent->type == REDIS_REPLY_PUSH);
         parent->element[task->idx] = r;
     }
     return r;
@@ -260,7 +261,8 @@ static void *createNilObject(const redisReadTask *task) {
         parent = task->parent->obj;
         assert(parent->type == REDIS_REPLY_ARRAY ||
                parent->type == REDIS_REPLY_MAP ||
-               parent->type == REDIS_REPLY_SET);
+               parent->type == REDIS_REPLY_SET ||
+               parent->type == REDIS_REPLY_PUSH);
         parent->element[task->idx] = r;
     }
     return r;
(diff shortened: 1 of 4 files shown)


michael-grunder:
Thanks for the PR.  I'll try to go over the changes in detail this week, but you do have an off-by-one/underflow error in your refactor of `seekNewline` when `len == 0`.

I refactored the logic a bit over on [my fork](https://github.com/michael-grunder/hiredis/tree/reader-updates) and it seems to fix the problem.

alexsmith1612:
Good catch on the `len == 0` case. I was trying to maintain both the general logic and style (in terms of not updating the argument variables themselves) of the current implementation and completely missed that case. Your refactored logic looks to me.

alexsmith1612:
@michael-grunder Any update on this?

If it's helpful, I can do any/all of:

- Cherry-pick your off-by-one fix michael-grunder/hiredis@f34f1794cc020f7eb97496d6b08851ace89bb3cf to my PR branch
- Directly fix my alexsmith1612/hiredis@36967ee3f0cf95954811f086d54ee473f0522e52 commit with a rebase/fixup
- Rebase this PR branch on top of the upstream master
- Split this branch/PR into multiple smaller PRs

michael-grunder:
Hi @alexsmith1612,

I'll get this merged tonight.  I figured I'd cherry-pick the fix onto the PR and then merge.

alexsmith1612:
Thank you. Let me know if there are any other changes you think should be made.

I cherry-picked your fix onto the PR.

michael-grunder:
Merged via d6a0b192b414dc261589d4c2d16577c2192b39c0, thanks!

I'm not totally sure why GitHub didn't automatically detect that, but all the commits are in.

Thanks again!

alexsmith1612:
Thanks for merging!

Regarding github not automatically detecting the PR as merged, my two thoughts are that the reason is one or both of the following:
- The rebased branch and commits that were merged to master were never actually on the branch associated with this PR (i.e. in my fork)
- The merge commit did not mention this PR number

Regardless, it doesn’t really make a difference.

PR #896, opened 2020-11-10
Free SSL object when redisSSLConnect fails

ayeganov:
# What Does This PR Do?

This set of changes removes a constant leak that occurs in `redisInitiateSSLWithContext`. Specifically, the issue is easily reproducible by trying to connect to an invalid IP address. 

1. The `SSL` object is created on line 383 in `redisInitiateSSLWithContext`:
    `SSL *ssl = SSL_new(redis_ssl_ctx->ssl_ctx);`

2. Which gets passed down to `redisSSLConnect`:
    `return redisSSLConnect(c, ssl)`

3. The `SSL_connect` function fails and the code falls through to to line 341 where the error string gets filled in:
    ```
    if (c->err == 0) {
        char err[512];
        if (rv == SSL_ERROR_SYSCALL)
            snprintf(err,sizeof(err)-1,"SSL_connect failed: %s",strerror(errno));
        else {
            unsigned long e = ERR_peek_last_error();
            snprintf(err,sizeof(err)-1,"SSL_connect failed: %s",
                    ERR_reason_error_string(e));
        }   
        __redisSetError(c, REDIS_ERR_IO, err);
    }
    ```

The function returns an error at this point, and the handle to `SSL` object is lost. Subsequent connections repeat this failure and start leaking large chunks of memory.

The fix itself is trivial - add the `SSL_free(ssl)` call before leaving the function.

### DIFF
--- a/ssl.c
+++ b/ssl.c
@@ -351,6 +351,7 @@ static int redisSSLConnect(redisContext *c, SSL *ssl) {
     }
 
     hi_free(rssl);
+    SSL_free(ssl);
     return REDIS_ERR;
 }
 


michael-grunder:
Merged, thanks!

ayeganov:
@michael-grunder Will this fix be available any time soon through a tag? Also, are you guys the ones who publish hiredis as a conan package, or is that done by someone else who is being nice?

michael-grunder:
I don't do any packing for Hiredis so I'm not sure who creates the conan package.

I think we're at a pretty good point to do v1.0.1 but I want to take another detailed look to make sure there aren't any breaking changes.

zandbelt:
I would really like to  see a new release with this fix because with out it v1.0.0 is not applicable in production environments.

PR #904, opened 2020-12-23
Add adapters/libhv

ithewei:
Build steps:
```
git clone https://github.com/ithewei/libhv
cd libhv
./configure && make && sudo make install

git clone -b libhv https://github.com/ithewei/hiredis
cd hiredis
mkdir build
cd build
cmake .. -DENABLE_EXAMPLES=ON
make
./examples/example-libhv
```

output:
```
Connected...
argv[end-1]: ./examples/example-libhv
Disconnected...
```

### DIFF
--- a/adapters/libhv.h
+++ b/adapters/libhv.h
@@ -1,0 +1,69 @@
+#ifndef __HIREDIS_LIBHV_H__
+#define __HIREDIS_LIBHV_H__
+
+#include <hv/hloop.h>
+#include "../hiredis.h"
+#include "../async.h"
+
+static void redisLibhvHandleEvents(hio_t* io) {
+    redisAsyncContext* context = (redisAsyncContext*)hevent_userdata(io);
+    int events = hio_events(io);
+    int revents = hio_revents(io);
+    if (context && (events & HV_READ) && (revents & HV_READ)) {
+        redisAsyncHandleRead(context);
+    }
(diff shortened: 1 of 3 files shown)


michael-grunder:
This PR looks good to me, but I can't find a great deal of information about `libhv` on the web, other than it has almost [5000 stars](https://github.com/ithewei/libhv) on GitHub.

@bosv Any thoughts?  I don't really see a downside to adding another adapter.  If we merge it I'll update the `Makefile` to so it can build the example as well.

adapters/libhv.h
```diff
@@ -0,0 +1,69 @@
...
+        return REDIS_ERR;
+    }
+    hevent_set_userdata(io, ac);
+
+    ac->ev.addRead  = redisLibhvAddRead;
+    ac->ev.delRead  = redisLibhvDelRead;
+    ac->ev.addWrite = redisLibhvAddWrite;
+    ac->ev.delWrite = redisLibhvDelWrite;
+    ac->ev.cleanup  = redisLibhvCleanup;
+    ac->ev.data     = io;
```
bjosv, line 65: The adapter is not supporting timeouts via `ac->ev.scheduleTimer`, but I guess that's ok.
It's only supported by 4 of the 9 current adapters.

bjosv:
I have tested the example and it works fine to build and run, so LGTM!

PR #906, opened 2020-12-29
Don't leak memory if an invalid type is set

rouzier:


michael-grunder:
This seems reasonable.  I'll test/merge it over the weekend.

rouzier:
Bump!

michael-grunder:
Merged, thanks!

PR #912, opened 2021-01-09
Add MSYS2 support

taozuhong:


bjosv (commented):
This is nice, but I believe Windows builds of hiredis are best supported via CMake.
There has been some work recently regarding debug info and openssl support, and it also builds with MinGW/MSYS2.

Would it be acceptable to use CMake instead?

PR #916, opened 2021-01-21
Fix crash bug during parse reply

bugwz:
**In the scenario of parsing certain strings, the parsing process will crash**

Crash occurs when parsing the string `*3\r\n$3\r\nSET\r\n$5\r\nhello\r\n$`, the reason is that in some scenarios, `r->pos` and `r->len` will be equal, leading to crash.

The related problem code is as follows:

```c
static int processBulkItem(redisReader *r) {
    redisReadTask *cur = r->task[r->ridx];
    void *obj = NULL;
    char *p, *s;
    long long len;
    unsigned long bytelen;
    int success = 0;

    p = r->buf+r->pos;
    s = seekNewline(p,r->len-r->pos);
    if (s != NULL) {
        p = r->buf+r->pos;
        bytelen = s-(r->buf+r->pos)+2; /* include \r\n */
        
       /* omit part of the code .... */
    }

    return REDIS_ERR;
}
```

- `_len` has the risk of overflow
- `len` may be 0

```c
/* Find pointer to \r\n. */
static char *seekNewline(char *s, size_t len) {
    int pos = 0;
    int _len = len-1;

    /* omit part of the code .... */
}

```

### DIFF
--- a/read.c
+++ b/read.c
@@ -123,13 +123,17 @@ static char *readBytes(redisReader *r, unsigned int bytes) {
 
 /* Find pointer to \r\n. */
 static char *seekNewline(char *s, size_t len) {
-    int pos = 0;
-    int _len = len-1;
+    size_t _len, pos = 0;
+
+    if (len == 0) {
+        return NULL;
+    }
 
     /* Position should be < len-1 because the character at "pos" should be
      * followed by a \n. Note that strchr cannot be used because it doesn't
      * allow to search a limited length and the buffer that is being searched
      * might not have a trailing NULL character. */
+    _len = len-1;
     while (pos < _len) {
         while(pos < _len && s[pos] != '\r') pos++;
         if (pos==_len) {
--- a/test.c
+++ b/test.c
@@ -547,6 +547,19 @@ static void test_reply_reader(void) {
     test_cond(ret == REDIS_ERR && reply == NULL);
     redisReaderFree(reader);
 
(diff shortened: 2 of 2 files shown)


michael-grunder:
Hi @bugwz, thanks for the PR.

I was planning on landing [this branch](https://github.com/michael-grunder/hiredis/tree/reader-updates) soon which makes modifications to the same part of the code, in addition to making it faster.

It actually looks like a similar off-by-one [fix](https://github.com/michael-grunder/hiredis/commit/f34f1794cc020f7eb97496d6b08851ace89bb3cf) is in that branch as well.

Let me get this merged and then verify that the bug is no longer present and also cherry-pick your test-case.

bugwz:
@michael-grunder copy that～

bjosv (commented):
Nice finding. Which arch/platform did you use to get the crash?
(Tried to run the testcase on v1.0.2 on a 64bit linux without "luck")

This works fine on master now btw.

In general maybe we should have more tests like the
`test("Works when a single newline (\\r\\n) covers two calls to feed: ")`
to cover more splitted readings during
`processLineItem(r)`, `processBulkItem(r)` and `processAggregateItem(r)`

michael-grunder:
Hi @bugwz I cherry-picked your test and it passes.

I had to do some git shenanigans as this required `cherry-pick -n` and setting you as the author in the ultimate commit.  As long as it looks good to you, I will get it merged.  (PR: #1092)

PR #919, opened 2021-01-26
Handle OOM during async command callback registration

bjosv:
Unless the callback is pushed for a regular command an assert in redisProcessCallbacks() will hit when the response arrives:
`/* No more regular callbacks and no errors, the context *must* be subscribed or monitoring. */
`

This change let the user get an early error instead, which is available in the async context directly.

This issue has been found in following [testcase](https://github.com/Nordix/hiredis-cluster/blob/master/tests/ct_out_of_memory_handling.c#L390).


zuiderkwast (approved):
LGTM

michael-grunder:
Merged, thanks!

PR #926, opened 2021-02-19
Add finalizer support to async Redis client

### LINKED ISSUE #925: async: callback lifecycle management
I'm trying to wrap hiredis (in async mode) from C++, and I'd like the callbacks to have associated state that is freed when no longer needed.

For non-pub/sub callbacks, this is easy. Hiredis will call each callback exactly once (when the reply is received, or when the connection is lost).

For pub/sub callbacks, this is not so easy. The pub/sub callback can be called zero times (if I subscribe to the same channel a second time, the old callback is not called any more), or an unbounded number of times (once for the initial "subscribe" / "psubscribe", once for each message, once for the final "unsusbcribe" / "punsubscribe").

Also, because I'm trying to wrap hiredis in a relatively transparent way, I don't even know whether a command is pub/sub or not, and I'd rather not replicate the parsing logic.

RFC: I'd like to add a second callback mode in which callbacks are always called at least once, and the last call is done with a NULL reply, which I could use to free the associated state.

Non-pub/sub callbacks would be called twice during normal operation: once with the actual reply, once with NULL (or just once with NULL if the context is being freed).

Pub/sub callbacks would be called at least once: if replacing a callback, the old one is called with NULL. Otherwise, the callback is called N times (once for "subscribe", once per message, once for "unsubscribe") and then again with NULL at the last unsubscribe.

I'd be happy to send a PR, let me know if you think this idea has merit.

tudor:
Also add libevent example that tests it and also demonstrates pubsub.

Fixes #925

### DIFF
--- a/async.c
+++ b/async.c
@@ -291,6 +291,21 @@ static void __redisRunCallback(redisAsyncContext *ac, redisCallback *cb, redisRe
         c->flags |= REDIS_IN_CALLBACK;
         cb->fn(ac,reply,cb->privdata);
         c->flags &= ~REDIS_IN_CALLBACK;
+        /* Always run finalizer when calling with NULL reply. */
+        if (reply == NULL && cb->finalizer != NULL) {
+            c->flags |= REDIS_IN_FINALIZER;
+            cb->finalizer(ac,cb->privdata);
+            c->flags &= ~REDIS_IN_FINALIZER;
+        }
+    }
+}
+
+static void __redisRunFinalizer(redisAsyncContext *ac, redisCallback *cb) {
+    redisContext *c = &(ac->c);
+    if (cb->fn != NULL && cb->finalizer != NULL) {
+        c->flags |= REDIS_IN_FINALIZER;
+        cb->finalizer(ac,cb->privdata);
+        c->flags &= ~REDIS_IN_FINALIZER;
     }
 }
 
@@ -362,7 +377,7 @@ static void __redisAsyncFree(redisAsyncContext *ac) {
 void redisAsyncFree(redisAsyncContext *ac) {
     redisContext *c = &(ac->c);
     c->flags |= REDIS_FREEING;
-    if (!(c->flags & REDIS_IN_CALLBACK))
+    if (!(c->flags & (REDIS_IN_CALLBACK | REDIS_IN_FINALIZER)))
         __redisAsyncFree(ac);
 }
 
@@ -406,11 +421,11 @@ void redisAsyncDisconnect(redisAsyncContext *ac) {
 
     /** unset the auto-free flag here, because disconnect undoes this */
     c->flags &= ~REDIS_NO_AUTO_FREE;
-    if (!(c->flags & REDIS_IN_CALLBACK) && ac->replies.head == NULL)
+    if (!(c->flags & (REDIS_IN_CALLBACK | REDIS_IN_FINALIZER)) && ac->replies.head == NULL)
         __redisAsyncDisconnect(ac);
 }
 
-static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply, redisCallback *dstcb) {
+static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply, redisCallback *dstcb, int* call_finalizer) {
     redisContext *c = &(ac->c);
     dict *callbacks;
     redisCallback *cb;
@@ -421,6 +436,7 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
 
     /* Custom reply functions are not supported for pub/sub. This will fail
      * very hard when they are used... */
+    *call_finalizer = 0;
     if (reply->type == REDIS_REPLY_ARRAY || reply->type == REDIS_REPLY_PUSH) {
         assert(reply->elements >= 2);
         assert(reply->element[0]->type == REDIS_REPLY_STRING);
@@ -451,8 +467,10 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
 
             /* If this is an unsubscribe message, remove it. */
             if (strcasecmp(stype+pvariant,"unsubscribe") == 0) {
-                if (cb->pending_subs == 0)
+                if (cb->pending_subs == 0) {
                     dictDelete(callbacks,sname);
+                    *call_finalizer = 1;
+                }
 
                 /* If this was the last unsubscribe message, revert to
                  * non-subscribe mode. */
@@ -469,6 +487,7 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
     } else {
         /* Shift callback for invalid commands. */
         __redisShiftCallback(&ac->sub.invalid,dstcb);
+        *call_finalizer = 1;
     }
     return REDIS_OK;
 oom:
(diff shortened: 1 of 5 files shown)


tudor:
@michael-grunder have you had a chance to take a look at this? Thank you!

michael-grunder:
Hi @tudor, sorry for the delay.

The logic looks good at first glance.  We have to be pretty careful about backward compatibility now that we've released v1.0.0, but I will test the logic in detail over the next couple of days.

tudor:
@michael-grunder I don't think there are any backwards compatibility issues. The only thing that changes in a potentially backwards incompatible way is `struct redisCallback` (its size and the offsets of the `pendingSubs` and `privdata` members), but that's not allocated explicitly by users, nor is it used in any inline functions (in fact, it probably shouldn't be in a public header file at all).

tudor:
I guess code that explicitly accesses `redisAsyncContext::replies` would break, but I don't imagine that was ever intended for public consumption.

kristjanvalur:
I was just about to open an issue asking for this yesterday.  I presume this is to support knowing when hiredis forgets the callback.
A couple of questions:

1. The finalizer seems necessary only for subsriptions, because regular callbacks are called once, and only once, right?
2. There is code in the subscription logic that _replaces_ subscription callbacks in the dicts, if a previous subscription exists.  Are the finalizers called in that case?  Should they?  Only if the `privdata` member is different?

tudor:
@kristjanvalur 

1. I chose to run the finalizer for regular callbacks as well, because that allows me to generically free state in the C++ code that sits on top of hiredis without having to know whether the callback is a subscription callback or not. If you don't want a finalizer, pass `NULL` :)
2. Similarly, I chose to always run the finalizer in the replacement case. Your old callback will never be called again (finalize it), and the new one will.

async.c
```diff
@@ -782,6 +809,7 @@ static int __redisAsyncCommand(redisAsyncContext *ac, redisCallbackFn *fn, void
             if (de != NULL) {
                 existcb = dictGetEntryVal(de);
                 cb.pending_subs = existcb->pending_subs + 1;
+                __redisRunFinalizer(ac,existcb);
```
kristjanvalur, line 812: Wouldn't it be prudent to check here if the contents of the callback is different from the replacement?  In particular the `privdata` member, but probably also the callback and finalizer, in case the caller uses dynamically created stubs and `NULL` for `privdata`.

kristjanvalur (changes requested):

async.c
kristjanvalur, line 812:
```C
if (existcb->privdata != cb.privdata)
  __redisRunFinalizer(ac, existcb);
```

tudor:
@kristjanvalur sorry for the long delay, I made the change you requested. PTAL?

kristjanvalur:
LGTM, although honestly, not much is happening in this repo.  I've got a bunch of PRs that have been stuck for months, including a bunch of defects/fixes.

tudor:
@michael-grunder PTAL? C++ programmers worldwide would thank you :)

async.h
```diff
@@ -42,9 +42,11 @@ struct dict; /* dictionary header is included in async.c */
 
 /* Reply callback prototype and container */
 typedef void (redisCallbackFn)(struct redisAsyncContext*, void*, void*);
+typedef void (redisFinalizerCallback)(struct redisAsyncContext*, void*);
 typedef struct redisCallback {
     struct redisCallback *next; /* simple singly linked list */
     redisCallbackFn *fn;
+    redisFinalizerCallback *finalizer;
     int pending_subs;
     void *privdata;
```
michael-grunder, line 51: I think we have to put the new member at the end of the structure,  as the current changes breaks ABI compatibility by inserting a member in the middle.

```suggestion
    int pending_subs;
    void *privdata;
    redisFinalizerCallback *finalizer;
```

Screen Shot 2021-08-08 at 7 09 28 PM

async.c
```diff
@@ -502,11 +521,12 @@ static int redisIsSubscribeReply(redisReply *reply) {
 
 void redisProcessCallbacks(redisAsyncContext *ac) {
     redisContext *c = &(ac->c);
-    redisCallback cb = {NULL, NULL, 0, NULL};
+    redisCallback cb = {NULL, NULL, NULL, 0, NULL};
```
michael-grunder, line 524: Since we're placing the new callback at the end.

```suggestion
    redisCallback cb = {NULL, NULL, 0, NULL, NULL};
```

examples/CMakeLists.txt
```diff
@@ -21,8 +21,10 @@ ENDIF()
 
 FIND_PATH(LIBEVENT event.h)
 if (LIBEVENT)
-    ADD_EXECUTABLE(example-libevent example-libevent)
+    ADD_EXECUTABLE(example-libevent example-libevent.c)
     TARGET_LINK_LIBRARIES(example-libevent hiredis event)
+    ADD_EXECUTABLE(example-libevent-pubsub example-libevent-pubsub.c)
+    TARGET_LINK_LIBRARIES(example-libevent-pubsub hiredis event)
```
michael-grunder, line 27: We should also update the `Makefile` with this new example, although I'm happy to do that after we merge the PR.

examples/example-libevent-pubsub.c
```diff
@@ -0,0 +1,169 @@
...
+        if (c->errstr) {
+            printf("errstr: %s\n", c->errstr);
+        }
+        return;
+    }
+    printf("argv[%s]: %s\n", (char*)privdata, reply->str);
+}
+
+void getFinalizer(redisAsyncContext *c, void *privdata) {
+    printf("Get finalizer called\n");
```
michael-grunder, line 62: Just to avoid the warning

```suggestion
    (void)privdata;
    printf("Get finalizer called\n");
```

examples/example-libevent-pubsub.c
```diff
@@ -0,0 +1,169 @@
...
+
+    if (sd->break_loop) {
+        sd->break_loop = 0;
+        redisLibeventEvents* e = c->ev.data;
+        event_base_loopbreak(e->base);
+    }
+}
+
+void subscribeFinalizer(redisAsyncContext *c, void *privdata) {
+    printf("Subscribe finalizer called\n");
```
michael-grunder, line 118:
```suggestion
    (void)privdata;
    printf("Subscribe finalizer called\n");
```

async.c
michael-grunder: `redisCallback cb = {0};` would be fine too.

michael-grunder (changes requested):
If you don't mind making those small changes I will get this merged tomorrow morning (West coast, PDT).

zuiderkwast:
@tudor are you still interested in finishing this?

tudor:
@zuiderkwast No, I'm not planning to work on this any more. I ended up having to parse the request and response (to know whether it's pub/sub or regular command) anyway for a different reason, in which case I can tell when the callback may be called multiple times.

kristjanvalur:
That's a shame, it looked like a worthwhile addition.  If there is interest, I can pick this up.

zuiderkwast:
@kristjanvalur I was going to pick it up myself, but you can do it if you want.

We want it for hiredis-cluster, so we can implement pubsub without actually looking at the commands sent by the user. We haven't fully decided if this is enough though.

zuiderkwast:
Hehe, I'm not sure I'm less rusty with hiredis internals. :-) I was just going to rebase and apply Michael's comments. I'll invite you for review.

zuiderkwast:
This PR doesn't correctly handle SUBSCRIBE with multiple channels with finalizer.

I've implemented this now in #1173. It's ready for review. Please have a look. (I'll try to get time to implement the RESET command too, and to allow calling commands while in monitor mode.)

I had to rewrite a lot, because a callback can be used for multiple channels, which can be usubscribed independenly. To keep track of the last reference to a callback (for calling the finalizer at the right time) I've added a reference counter in the callback struct.

Issue #927, opened 2021-03-13
[hiredis_ssl] Add support for skipping certificate verification


sewenew:
**redis-cli 6.2** already supports skipping certificate verification and setting default verification path: [#8416](https://github.com/redis/redis/commit/be83bb13a8eaad68b7580b95c696f2554cf7100e#). However, it seems that hiredis_ssl doesn't catch up with this feature.

Is there a plan to port this feature to hiredis_ssl? Thanks!

Regards

michael-grunder:
Closing as #1085 was merged

PR #928, opened 2021-03-16
set default SSL verification path

adobeturchenko:
Adding missing default verify path for client certificates.

Issue causing connection fail if certificate is not specified in connection settings (it's optional so it doesn't need to be specified).
Fixes https://github.com/redis/hiredis/issues/927 however issue is not skipping verification but setting default path.


michael-grunder:
I'm going through all of our open PRs since I'm putting together a new version soon.

I'm inclined to merge this as the feature adds quite a lot of convenience (e.g. to connect to a cloud provider Redis endpoint) without having to do all of the work manually via `redisInitiateSSL`.

The only real question I have is whether we need to put this behind a `redisConnectWithOptions` option as it would change Hiredis' current behavior by default.

michael-grunder:
Merged, thanks!

We can bikeshed whether we need to put it behind an option when putting together the next release.

PR #930, opened 2021-03-26
Bugfix: Support failed async connects on windows.

kristjanvalur:
On Windows, a failed non-blocking connect() must be resolved by getting the socket error using getsockopt().  In fact, this
is the recommended way on posix too, not to repeat a call to connect(), but call getsockopt until it succeeds or fails.
This PR adds error translation to getsockopt(), adds an edge case for winsock (WSAEINVAL) and enables getsockopt() processing when connect() indicates that it is prudent. 
This allows async connects to fail on windows with the proper error code.


kristjanvalur:
Please not that the _correct_ way to do a non-blocking `connect()` call is make a **single** call to `connect()`, then subsequently do only `getsockopt()` to query the socket status.  Otherwise, `connect()` would have been endowed with a **EWOULDBLOCK** error like the other calls designed to be non-blocking.   This is the reason for the multiple platform-dependent way to check for a failing non-blocking connect, there wasn't really a specification for how second call should behave.  
It would be cleaner to have a separate function to _initiate_ a connect and another to _check_ it subsequently, but I didn't want to make the patch overly complicated and risky.

kristjanvalur:
Pinging this.

kristjanvalur:
@michael-grunder , could we get this merged?

michael-grunder:
I cherry-picked your commits to keep the git history clean.

kristjanvalur:
No problem.  You can also ask me to rebase these PRs before merging, they've been lingering here for a long time and had Main merged in quite often.

PR #931, opened 2021-03-26
Stability: Support calling redisAsyncCommand and redisAsyncDisconnect from the onConnected callback

kristjanvalur:
There is nothing prevengint redisAsyncDisconnect() from being called from the onConnect() callback, and in fact, an application may decide that it doesn't want the connection.
However, redisAsyncDisconnect() will delete the context, before other processing involving the context can finish, when handling the original connection.  This causes a crash.

This PR adds the usual callback flag when calling the connect/disconnect callbacks.  Any pending disconnect or free is then handled by the __redisAsyncHandleConnect() handler.
It is now safe to delete the collection from the onConnect callback. 

Additionally, the REDIS_CONNECTED flag is now set _before_ the onConnect() callback is called, which makes the disconnect logic work correctly from with the callback.

In addition, any other redis commands can be called from the on_connected, such as redisAsyncCommand(), to directly schedule commands for execution.

The PR includes regression tests for both cases.

### DIFF
--- a/async.c
+++ b/async.c
@@ -303,6 +303,34 @@ static void __redisRunPushCallback(redisAsyncContext *ac, redisReply *reply) {
     }
 }
 
+static void __redisRunConnectCallback(redisAsyncContext *ac, int status)
+{
+    if (ac->onConnect) {
+        if (!(ac->c.flags & REDIS_IN_CALLBACK)) {
+            ac->c.flags |= REDIS_IN_CALLBACK;
+            ac->onConnect(ac, status);
+            ac->c.flags &= ~REDIS_IN_CALLBACK;
+        } else {
+            /* already in callback */
+            ac->onConnect(ac, status);
+        }
+    }
+}
+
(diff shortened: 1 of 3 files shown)


kristjanvalur:
Simplified the change by leveraging the `REDIS_IN_CALLBACK`` and `REDIS_FREEING` flags when called for the connect/disconnect callbacks, and using helper functions to call them.

kristjanvalur:
Any news on this?

kristjanvalur:
It would be awesome to get any kind of feedback on this pull request.

kristjanvalur:
Rebased on top of main.  Added regression test.  Drop `const` qualifier on the redisAsyncContext in the connection callback, so that it can call api functions without warnings.

kristjanvalur:
Would be great to get a review on this.  Adds parity to the `onConnected` callback so that other APIs can be called directly from it.

kristjanvalur:
🎉

PR #932, opened 2021-03-27
Polling adapter and example

kristjanvalur:
An adaptor to use redisAsyncContext using simple polling.  Useful in frameworks where an IO loop isn't present.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 # This file is released under the BSD license, see the COPYING file
 
 OBJ=alloc.o net.o hiredis.o sds.o async.o read.o sockcompat.o
-EXAMPLES=hiredis-example hiredis-example-libevent hiredis-example-libev hiredis-example-glib hiredis-example-push
+EXAMPLES=hiredis-example hiredis-example-libevent hiredis-example-libev hiredis-example-glib hiredis-example-push hiredis-example-poll
 TESTS=hiredis-test
 LIBNAME=libhiredis
 PKGCONFNAME=hiredis.pc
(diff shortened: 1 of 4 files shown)


michael-grunder:
Hi, @kristjanvalur thanks for the PR.  I'll take a look at the logic this week.

kristjanvalur:
Right.  This is extracted from some work I did to make this jive from within UnrealEngine.

michael-grunder:
I rebased your PR into a single commit and made a few (mostly formatting) changes/suggestions in [a branch on my fork](https://github.com/michael-grunder/hiredis/tree/polling-adapter).  Mostly it's just to maintain hiredis' coding style.

The only actually changed logic:

1.  Move timeval -> double (and vice versa) conversion into helpers.
2. Create `REDIS_POLL_XXX` defines to signal which events were processed in `redisPollTick`

If you're OK with the changes feel free to apply the patch as whitespace shouldn't make me the author.

One question I did have, was whether we may wish to handle the case where `select` returns -1 and sets errno?

kristjanvalur:
Great, I'll have a look, and think about the question

kristjanvalur:
Ok, yes, IMHO we should check for -1 out of select() .  Since the socketcompat.h aldready has a helper for poll(), I can change the code to use poll and we get the platform compatible error handling for free.

kristjanvalur:
I've cherry picked the initial async tests I had added in pr [#948](https://github.com/redis/hiredis/pull/948)
At the time, there were _no_ async tests in test.c.  Now I see that there is one test, using libevent.  The advantage of using polling adapter is that there is no external dependency and no mysterious external event loop to deal with.  It also requires no special #ifdef for that reason.

#948 already has this test, and tests for a number of other issues that I fixed and are rolled up in that PR.

kristjanvalur:
This PR would be highly appreciated.  For one thing, writing unit-tests using the polling method is much simpler, since it doesn't require any additional scaffolding or external libraries.

If there is anything I need to fix up (style, formatting, collapsing changes, etc) I'm happy to do so.

michael-grunder:
If you wouldn't mind rebasing off of master and squashing the polling adapter commits and async test commits where possible, that would be great.  Since I have a commit in the middle a full squash is probably not practical.

[I took a crack at it on my fork](https://github.com/michael-grunder/hiredis/tree/kristjanvalur-pr3-rebase), and thankfully it wasn't too painful.

kristjanvalur:
Its now been cleaned and collapsed into a couple of distinct commits.  I'm unsure about whatever formatting concerns you might have, what style rules should be followed.

PR #935, opened 2021-04-08
Bugfix: Ignore timeout callback from a successful connect (fixes #945)

### LINKED ISSUE #945: Stray timeouts from async connect can interrupt commands
See Pull Request #935

kristjanvalur:
If `redisAsyncConnect` is done with a connect_timeout, and subsequently a redisAsyncCommand() happens but with no command_timeout, the connect timeout can still arrive, since it is never cancelled.  This fix prevents such a belated timeout from cancelling a command.

### DIFF
--- a/async.c
+++ b/async.c
@@ -690,9 +690,17 @@ void redisAsyncHandleTimeout(redisAsyncContext *ac) {
     redisContext *c = &(ac->c);
     redisCallback cb;
 
-    if ((c->flags & REDIS_CONNECTED) && ac->replies.head == NULL) {
-        /* Nothing to do - just an idle timeout */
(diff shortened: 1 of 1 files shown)


kristjanvalur:
fixes #945

michael-grunder:
Thanks for the fix, merged.

kristjanvalur:
Great.  You'll find a regression test for this issue in PR #948  (those async tests depend on the "polling adapter" from pr #932

michael-grunder:
Sounds good.  

I plan on merging the adapter too but wanted to run it locally a bit so I have a good enough understanding of it to answer questions about it.

PR #936, opened 2021-04-08
Fix int/size_t discrepancies, warnings on windows builds.

kristjanvalur:
A lot of discrepancies with int vs size_t were visible when building on windows.
This PR fixes those.  At least one case was an error, where an attempt was made to negate an unsigned value.


hiredis.c
```diff
@@ -516,7 +516,7 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
 
     hi_free(curargv);
     *target = cmd;
-    return totlen;
+    return (int)totlen;
```
kristjanvalur, line 519: These commands really should return a `ssize_t` result to be consistent with the use of `size_t`, but I didn't want to modify their signature since they are a public API

kristjanvalur, line 519: These commands really should return a `ssize_t` result to be consistent with the use of `size_t`, but I didn't want to modify their signature since they are a public API

net.c
```diff
@@ -457,7 +457,7 @@ static int _redisContextConnectTcp(redisContext *c, const char *addr, int port,
             }
 
             for (b = bservinfo; b != NULL; b = b->ai_next) {
-                if (bind(s,b->ai_addr,b->ai_addrlen) != -1) {
+                if (bind(s,b->ai_addr,(socklen_t)b->ai_addrlen) != -1) {
```
kristjanvalur, line 460: Windows has ai_addren as `size_t` rather than `socklen_t`.  explicitly cast it to silence warning.

net.c
```diff
@@ -600,13 +600,13 @@ int redisContextConnectUnix(redisContext *c, const char *path, const struct time
 
     c->flags |= REDIS_CONNECTED;
     return REDIS_OK;
+oom:
```
kristjanvalur, line 603: Fix error about unused label (windows)

read.c
```diff
@@ -213,7 +216,7 @@ static int string2ll(const char *s, size_t slen, long long *value) {
     if (negative) {
         if (v > ((unsigned long long)(-(LLONG_MIN+1))+1)) /* Overflow. */
             return REDIS_ERR;
-        if (value != NULL) *value = -v;
+        if (value != NULL) *value = -(long long)v;
```
kristjanvalur, line 219: This is a bug!.  The negation of an unsigned is a no-op.  It must first be cast to a signed int.

read.c
```diff
@@ -536,7 +540,7 @@ static int processAggregateItem(redisReader *r) {
             if (r->fn && r->fn->createArray)
                 obj = r->fn->createArray(cur,elements);
             else
-                obj = (void*)(long)cur->type;
+                obj = (void*)(intptr_t)cur->type;
```
kristjanvalur, line 543: Must use an integral type of the same size as pointer to avoid the (sensible) warning here.

kristjanvalur:
Any news here?

kristjanvalur:
In the meantime, there have been some slight fixes upstream.  Curiosly, the authors have decided to use `long long` in places in stead of the more standard `size_t` for reasons unknown.  It should be best to stick to `ssize_t` since this better reflects the memory model supported.

hiredis.c
kristjanvalur: Since then, have updated to reflect the new api which returns `long long`.  Why that was chosen instead of `size_t` or `ssize_t` is unclear.

kristjanvalur:
This PR fixes a lot of internal type discrepancies within functions, using `size_t` and `ssize_t` where appropriate.
It _does_ also contain inline comments such as 
```C
 return (long long) totlen; /* api really should use ssize_t */
```
I'll be happy to remove those, since they reflect my personal opinion :)

kristjanvalur:
Moved the discovered **bug**/bugfix into its separate commit for easy view

PR #937, opened 2021-04-08
Fix adapters/libevent.h compilation for 64-bit Windows

pbtummillo:
Where SOCKET is a 64-bit unsigned integer.

### DIFF
--- a/adapters/libevent.h
+++ b/adapters/libevent.h
@@ -50,7 +50,7 @@ static void redisLibeventDestroy(redisLibeventEvents *e) {
     hi_free(e);
 }
 
-static void redisLibeventHandler(int fd, short event, void *arg) {
+static void redisLibeventHandler(evutil_socket_t fd, short event, void *arg) {
     ((void)fd);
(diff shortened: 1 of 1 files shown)


bjosv (approved):
I can verify that the libevent adapter gives the warnings:
```
..\hiredis\adapters/libevent.h(102,38): warning C4028: formal parameter 1 different from declaration
..\hiredis\adapters/libevent.h(171,76): warning C4028: formal parameter 1 different from declaration
```
without this change (on a Win10 via virtualbox 64bit).

When using an `evutil_socket_t` it matches the signature of [event_callback_fn](http://www.wangafu.net/~nickm/libevent-2.0/doxygen/html/event_8h.html#aed2307f3d9b38e07cc10c2607322d758) where the `fd` argument is an [intptr_t](https://github.com/libevent/libevent/blob/master/include/event2/util.h#L308) on windows.
Works fine on Linux as well.

michael-grunder:
Thanks, merged!

Issue #956, opened 2021-06-03
【fuzz】缓冲区溢出 


zhangtaoXT5:
1、python infra/helper.py build_fuzzers --sanitizer address hiredis
2、python infra/helper.py run_fuzzer hiredis format_command_fuzzer -rss_limit_mb=0

==13==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000024715 at pc 0x00000054f1bb bp 0x7ffc63cf1490 sp 0x7ffc63cf1488
READ of size 1 at 0x602000024715 thread T0
SCARINESS: 12 (1-byte-read-heap-buffer-overflow)
    #0 0x54f1ba in redisvFormatCommand /src/hiredis/hiredis.c:231:11
    #1 0x54f45e in redisFormatCommand /src/hiredis/hiredis.c:460:11
    #2 0x54d599 in LLVMFuzzerTestOneInput /src/hiredis/format_command_fuzzer.c:51:5
    #3 0x458241 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:556:15
    #4 0x457985 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:470:3
    #5 0x459a57 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:698:19
    #6 0x45a4d5 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:832:5
    #7 0x4494ae in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:826:6
    #8 0x471c82 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19:10
    #9 0x7ff76646882f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
[...]

PR #957, opened 2021-06-03
fix heap-buffer-overflow

zhangtaoXT5:
fix heap-buffer-overflow in https://github.com/redis/hiredis/issues/956

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -477,8 +477,8 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
 
             touched = 1;
             c++;
-	    if (*c == '\0')
-                 break;
+        if (*c == '\0')
+            break;
         }
         c++;
     }
--- a/hiredis.c
+++ b/hiredis.c
@@ -477,6 +477,8 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
 
             touched = 1;
             c++;
+	    if (*c == '\0')
+                 continue;
         }
         c++;
     }
--- a/hiredis.c
+++ b/hiredis.c
@@ -477,8 +477,8 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
 
             touched = 1;
             c++;
-        if (*c == '\0')
-            break;
+            if (*c == '\0')
+                break;
         }
         c++;
     }
--- a/hiredis.c
+++ b/hiredis.c
@@ -478,7 +478,7 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
             touched = 1;
             c++;
 	    if (*c == '\0')
-             continue;
+                 break;
         }
         c++;
     }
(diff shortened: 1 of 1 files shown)


hiredis.c
```diff
@@ -477,6 +477,8 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
 
             touched = 1;
             c++;
+	    if (*c == '\0')
```
kristjanvalur, line 480: wrong indentation

michael-grunder:
Hi,

I'm going through PRs in preparation for a new release.  I see what you're fixing here, but can you provide an example format string where we would currently read past the end of the buffer?

I thing a regression test may be worthwhile.

hiredis.c
```diff
@@ -477,6 +477,8 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
...
+                 continue;
```
kristjanvalur, line 481:
```suggestion
             continue;
```

bjosv:
I ran the fuzzing tests described in the issue to be able to propose a testcase.
I have pinpointed the problem to the format specifier parser and will push an alternative fix for this, together with the testcases.

michael-grunder:
Closing in favor of #1097.  Thanks for finding and reporting the issue though!

kristjanvalur:
I contest closing this.  Incrementing a zero terminated string twice, without checking that there is indeed a zero in between increments, is dangerous.  One should always check for zeroes, even if one strongly believes that the string is correctly constructed.

hiredis.c
```diff
@@ -477,6 +477,8 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
 
             touched = 1;
             c++;
+	    if (*c == '\0')
+             continue;
```
kristjanvalur, line 481: I'd change this to `break` to make the intention more explicit.

gzur:
I am a professional programmer and I approve of this message.

michael-grunder:
@kristjanvalur I see your point.  It certainly doesn't hurt anything to apply both fixes.

hiredis.c
michael-grunder, line 481:
```suggestion
	    if (*c == '\0')
                 break;
```

hiredis.c
```diff
@@ -477,6 +477,8 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
 
             touched = 1;
             c++;
+	    if (*c == '\0')
+                 break;
```
michael-grunder, line 481:
```suggestion
        if (*c == '\0')
            break;
```

hiredis.c
```diff
@@ -477,6 +477,8 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
 
             touched = 1;
             c++;
+        if (*c == '\0')
+            break;
```
michael-grunder, line 481:
```suggestion
            if (*c == '\0')
                break;
```

PR #960, opened 2021-06-08
Fix some undefined behaviour

jengab:
- redisSSLContextError must always be initialized at defintion,
  otherwise when SSL connect succeeds it may not be assigned to a valid error.
  Thus the memory trash remains in the variable, which may sign a misleading error.


michael-grunder:
I cherry-picked your commit and merged via #1091

PR #961, opened 2021-06-08
Fix ProtocolError

mtuleika-appcast:
This PR fixes an issue with incorrect behaviour of redisBufferWrite after merging SSL support.

The following logic was moved from redisBufferWrite to rawWrite in [commit](https://github.com/redis/hiredis/commit/0c1454490669f3c95e3c8b0ac6a83582d14e30e0#diff-b61215f4090069fa2ec48b946a1f19df25f63b1eb9e35d5bf71bd9128ae14498R830)
```
if ((errno == EAGAIN && !(c->flags & REDIS_BLOCK)) || (errno == EINTR)) {
    /* Try again later */
} else {
    __redisSetError(c, REDIS_ERR_IO, NULL);
    return -1;
}
```

The "Try again later" thing stopped working after that commit. This produces runtime errors in hiredis-rb after upgrading its hiredis to the latest version.

To reproduce the error:
1. git clone https://github.com/appcast-inc/hiredis-rb.git (my fork)
2. git checkout 66d0807b26b807012193fa5d08c9944d9d982f01 (undo the commit which fixes the error above)
3. bundle install && bundle exec rake compile
4. and finally, run several times bundle exec rake test

And you will get a runtime error!

Reach me out if you have any questions.

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -974,7 +974,12 @@ int redisBufferWrite(redisContext *c, int *done) {
     if (sdslen(c->obuf) > 0) {
         ssize_t nwritten = c->funcs->write(c);
         if (nwritten < 0) {
-            return REDIS_ERR;
+            if ((errno == EWOULDBLOCK && !(c->flags & REDIS_BLOCK)) || (errno == EINTR)) {
+                /* Try again later */
+            } else {
+                __redisSetError(c, REDIS_ERR_IO, NULL);
+                return REDIS_ERR;
+            }
         } else if (nwritten > 0) {
             if (nwritten == (ssize_t)sdslen(c->obuf)) {
                 sdsfree(c->obuf);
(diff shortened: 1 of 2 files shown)


YegorZdanovich:
@michael-grunder hello, any change this PR can be merged? or maybe some additional work needed? 
thank you

michael-grunder:
Apologies, I will test and merge this weekend.

michael-grunder:
Hi @YegorZdanovich 

I need to play around with this a bit more to completely think through the right place to catch `EAGAIN`.  The reason it's a bit tricky is that our `write` function is now abstracted behind a function pointer.

```c
/* This will likely continue to work for any reasonable `write` pointer, but there is no guarantee 
   `c->funcs->write` will set `errno` as `write` and `send` do. */
ssize_t nwritten = c->funcs->write(c);
if (nwritten < 0) {
        /* Try again later */
    } else {
        __redisSetError(c, REDIS_ERR_IO, NULL);
        return REDIS_ERR;
    }
}
```

We may need to use a special return value from `c->funcs->write` to specify such a condition generically.

michael-grunder:
Hi, @mtuleika-appcast can you take a look at #1106.

PR #974, opened 2021-08-16
Corrections for event library adapters

bjosv:
Fix the unused parameter warning in the `libev` adapter.
This is triggered when a user is building the adapter with Clang and having e.g `-Wextra -Werror` warnings enabled,
but it seems to work fine with GCC. Verified for both `-DEV_MULTIPLICITY=0` and `-DEV_MULTIPLICITY=1`

Also correcting a CMake warning when building the example for `libevent`.

Details:
```
CC=clang make CFLAGS="-Wunused-parameter -Werror" clean examples

./adapters/libev.h:122:31: error: unused parameter 'loop' [-Werror,-Wunused-parameter]
in
static void redisLibevTimeout(EV_P_ ev_timer *timer, int revents) {
$ clang --version
clang version 10.0.0-4ubuntu1
```
```
cmake -DENABLE_EXAMPLES=ON ..
CMake Warning (dev) at examples/CMakeLists.txt:24 (ADD_EXECUTABLE):
  Policy CMP0115 is not set: Source file extensions must be explicit. 
```

### DIFF
--- a/adapters/libev.h
+++ b/adapters/libev.h
@@ -66,8 +66,9 @@ static void redisLibevWriteEvent(EV_P_ ev_io *watcher, int revents) {
 
 static void redisLibevAddRead(void *privdata) {
     redisLibevEvents *e = (redisLibevEvents*)privdata;
+#if EV_MULTIPLICITY
     struct ev_loop *loop = e->loop;
-    ((void)loop);
+#endif
     if (!e->reading) {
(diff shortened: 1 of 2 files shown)


zuiderkwast (approved):
The loop variable really looks unused. It can be worth commenting in every place that EV_A_ is expecting the loop variable.

PR #977, opened 2021-08-27
Removed unused import

bjornleffler:
This line causes a build breakage in environments that only expose some (public?) OpenSSL header files.

I'm not an expert in the area. Just noticed it and submitting a simple fix for it.

### DIFF
--- a/ssl.c
+++ b/ssl.c
@@ -44,7 +44,6 @@
 #endif
 
 #include <openssl/ssl.h>
-#include <openssl/err.h>
 
 #include "win32.h"
 #include "async_private.h"


bjornleffler:
Friendly ping on this.

zuiderkwast (approved):
LGTM

Apparently, this doesn't break anything. (A missing include would have caused a compile error, but we build with SSL in CI, so it would have been detected.)

bjosv (commented):
This will probably fail in CI when lifted, we use -Werror there now.
Removing the include from master gives following warnings for me:
```
hiredis/ssl.c:326:5: warning: implicit declaration of function ‘ERR_clear_error’ [-Wimplicit-function-declaration]
  326 |     ERR_clear_error();
hiredis/ssl.c:345:31: warning: implicit declaration of function ‘ERR_peek_last_error’ [-Wimplicit-function-declaration]
  345 |             unsigned long e = ERR_peek_last_error();
      |                               ^~~~~~~~~~~~~~~~~~~
hiredis/ssl.c:347:21: warning: implicit declaration of function ‘ERR_reason_error_string’ [-Wimplicit-function-declaration]
  347 |                     ERR_reason_error_string(e));
      |                     ^~~~~~~~~~~~~~~~~~~~~~~
hiredis/ssl.c:346:62: error: format ‘%s’ expects argument of type ‘char *’, but argument 4 has type ‘int’ [-Werror=format=]
  346 |             snprintf(err,sizeof(err)-1,"SSL_connect failed: %s",
      |                                                             ~^
      |                                                              |
      |                                                              char *
      |                                                             %d
  347 |                     ERR_reason_error_string(e));
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~                
      |                     |
      |                     int
```

zuiderkwast:
Oh... @bjosv so this PR is should instead be rejected then?

bjosv:
@zuiderkwast I believe so.
@bjornleffler Interesting, err.h is usually included in the required `openssl-devel` package or via Brews `openssl` on Mac, Windows I dont know. Which OS/distribution are you using? and which build system (Make or via CMake)?

michael-grunder:
Going to close this, as it looks like we do sometimes need the header.

PR #822, opened 2020-05-23
redisAsyncAppend family of commands

michael-grunder:
This PR contains the redisAsyncAppend* family of commands and type fixes by @oshamash

See #638 for the original PR and discussion.

PR #983, opened 2021-09-29
Hiredis: Advanced async-handling

oshamash:
+ hiredis.c: Convert redisSetError to take format and va_args
+ dict.*   : Fixed generic hash function -> cannot really take signed number
+ async.*  : Valdup function now handles OOM properly
             KeyCompare function uses proper sized integers
             nextArgument - change to use proper str-int conversion function
             Added redisAsyncAppend* function family to support buffer
                                                 optimizations in async-mode
             Added proper error handling on OOM errors [inc buffer rollback]

Review fixes

+ hiredis.c : Reverted back to memcpy(), will test *printf in different time
+ aysnc.c   : Minor changes for better readability

Handle unlikely/pathological OOM conditon cleanly.

More OOM checks

### DIFF
--- a/async.c
+++ b/async.c
@@ -52,36 +52,35 @@
 #define assert(e) (void)(e)
 #endif
 
-/* Forward declarations of hiredis.c functions */
-int __redisAppendCommand(redisContext *c, const char *cmd, size_t len);
-void __redisSetError(redisContext *c, int type, const char *str);
+/* Forward declaration of function in hiredis.c */
+void __redisSetError(redisContext *, int, const char *, ...);
+int __redisAppendCommand(redisContext *, const char *, size_t);
 
 /* Functions managing dictionary of callbacks for pub/sub. */
 static unsigned int callbackHash(const void *key) {
-    return dictGenHashFunction((const unsigned char *)key,
-                               sdslen((const sds)key));
+    return dictGenHashFunction(key, sdslen((const sds) key));
 }
 
 static void *callbackValDup(void *privdata, const void *src) {
-    ((void) privdata);
     redisCallback *dup;
+    ((void) privdata);
 
-    dup = hi_malloc(sizeof(*dup));
-    if (dup == NULL)
-        return NULL;
+    if (NULL != (dup = hi_malloc(sizeof(*dup))))
+        memcpy(dup, src, sizeof(*dup));
 
-    memcpy(dup,src,sizeof(*dup));
     return dup;
 }
 
 static int callbackKeyCompare(void *privdata, const void *key1, const void *key2) {
-    int l1, l2;
+    size_t l1, l2;
     ((void) privdata);
 
     l1 = sdslen((const sds)key1);
     l2 = sdslen((const sds)key2);
+
     if (l1 != l2) return 0;
-    return memcmp(key1,key2,l1) == 0;
+
+    return  0 == memcmp(key1, key2, l1);
 }
 
 static void callbackKeyDestructor(void *privdata, void *key) {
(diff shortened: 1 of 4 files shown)


oshamash:
@michael-grunder I took #822 (from #638) and decided to match to upstream, I think it's about time community will have access to ASYNC optimizations using the native hiredis client 😄

async.c
```diff
@@ -317,22 +321,20 @@ static void __redisAsyncFree(redisAsyncContext *ac) {
...
 
-    /* Run subscription callbacks with NULL reply */
-    if (ac->sub.channels) {
-        dictInitIterator(&it,ac->sub.channels);
-        while ((de = dictNext(&it)) != NULL)
-            __redisRunCallback(ac,dictGetEntryVal(de),NULL);
-
-        dictRelease(ac->sub.channels);
-    }
+    #define RUN_SUB_CALLBACKS(cb_dict) \
```
oshamash, line 324: Wanted to have the logic for iteration matching scenarios for future update cases

async.c
```diff
@@ -107,16 +106,10 @@ static redisAsyncContext *redisAsyncInitialize(redisContext *c) {
...
-    patterns = dictCreate(&callbackDict,NULL);
-    if (patterns == NULL)
-        goto oom;
-
-    ac = hi_realloc(c,sizeof(redisAsyncContext));
-    if (ac == NULL)
+    /* Allocate memory at the start - OOM on failure */
+    if (NULL == (channels = dictCreate(&callbackDict,NULL)) ||
+        NULL == (patterns = dictCreate(&callbackDict,NULL)) ||
+        NULL == (ac = hi_realloc(c,sizeof(redisAsyncContext))))
```
oshamash, line 112: this is a convention I'm fond of, I can revert, but I like that all allocations are in a block

async.c
```diff
@@ -472,6 +476,7 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
     }
     return REDIS_OK;
 oom:
+    sdsfree(sname);
```
oshamash, line 479: `sdsfree` handles null-scenario

async.c
```diff
@@ -707,8 +712,7 @@ void redisAsyncHandleTimeout(redisAsyncContext *ac) {
     }
 
     if (!c->err) {
-        __redisSetError(c, REDIS_ERR_TIMEOUT, "Timeout");
-        __redisAsyncCopyError(ac);
+        __redisAsyncSetError(ac, REDIS_ERR_TIMEOUT, "Timeout");
```
oshamash, line 715: cut usage short with new `__redisAsyncSetError`

async.c
```diff
@@ -735,17 +739,17 @@ static const char *nextArgument(const char *start, const char **str, size_t *len
         if (p == NULL) return NULL;
     }
 
-    *len = (int)strtol(p+1,NULL,10);
+    *len = (size_t) strtoull(p+1,NULL,10);
```
oshamash, line 742: safety first

michael-grunder:
I'm going to merge this but first I need to make sure that the change to `dictGenHashFunction` doesn't somehow cause pain in upstream Redis.

I think it's OK but we've had this kind of symbol conflict before (it's why all the sds symbols in hiredis are prefixed with `hi_`).

async.c
```diff
@@ -317,22 +321,20 @@ static void __redisAsyncFree(redisAsyncContext *ac) {
...
-            __redisRunCallback(ac,dictGetEntryVal(de),NULL);
-
-        dictRelease(ac->sub.channels);
-    }
+    #define RUN_SUB_CALLBACKS(cb_dict) \
+        if (cb_dict) { \
+            dictInitIterator(&it,cb_dict); \
+            while (NULL != (de = dictNext(&it))) \
+                pcb = dictGetEntryVal(de); \
+                if (pcb) __redisRunCallback(ac,pcb,NULL); \
```
michael-grunder, line 329: Need a block here.

```suggestion
            while (NULL != (de = dictNext(&it))) { \
                pcb = dictGetEntryVal(de); \
                if (pcb) __redisRunCallback(ac,pcb,NULL); \
            } \
```

oshamash: @michael-grunder 👍 good catch, will amend now

Issue #718, opened 2019-09-30
hiredis client connect is alive but hang at redisBufferRead read , actually the tcp socket is closed.


zhungdongmei:
This is different with https://github.com/redis/hiredis/issues/593.

I create a thread to monitor redis command .Have a hiredis client with monitor command. And at long times the monitor cann't accept any command(currently I don't know why .May be some hardware problem). 
tracing the progress I find the thread is alive But redis client is closed : 
1. through redis command **client list** 。It has no client connect with monitor
2. Program received signal SIGPIPE, Broken pipe.  when read. But redis connect is no any error.

pstack thread info:
Thread 64 (Thread 0x7f8c34d71700 (LWP 1480741)):
#0  0x00007f8c38e096fd in read () from /lib64/libpthread.so.0
#1  0x000055dac621a2bb in redisBufferRead (c=c@entry=0x7f8c180008e0) at hiredis.c:788
#2  0x000055dac621a6c2 in redisGetReply (c=0x7f8c180008e0, reply=0x7f8c34d70e90) at hiredis.c:874

gdb the redisBufferRead:
Breakpoint 1, redisBufferRead (c=c@entry=0x7f8c2406da50) at hiredis.c:788
788	    nread = read(c->fd,buf,sizeof(buf));
(gdb) p *c
$3 = {err = 0, errstr = '\000' <repeats 127 times>, fd = 46, flags = 3, obuf = 0x7f8c2402cce3 "", reader = 0x7f8c2406d3b0, 
  connection_type = REDIS_CONN_TCP, timeout = 0x7f8c2400ef90, tcp = {host = 0x7f8c24036ea0 "127.0.0.1", source_addr = 0x0, port = 6379}, 
  unix_sock = {path = 0x0}}
(gdb) n

Program received signal SIGPIPE, Broken pipe.
[Switching to Thread 0x7f8c36d75700 (LWP 1480737)]
0x00007f8c38e09bfb in send () from /lib64/libpthread.so.0
(gdb) c
Continuing.

Program received signal SIGPIPE, Broken pipe.
0x00007f8c38e09bfb in send () from /lib64/libpthread.so.0
(gdb) c
Continuing.
[Switching to Thread 0x7f8c35572700 (LWP 1480740)]

Breakpoint 2, redisBufferRead (c=c@entry=0x7f8c2406da50) at hiredis.c:785
785	    if (c->err)
(gdb) p *c
[...]

Issue #593, opened 2018-05-07
hiredis client hang at redisBufferRead, with timeout option and keepalive option set


Gwan-He:
Here is how the client program use hiredis to connect redis servers:
![image](https://user-images.githubusercontent.com/11644714/39692954-01bb257c-5215-11e8-9c5d-e8cb5191b610.png)
The client program callls the redisSetTimeout and redisEnableKeepAlive to set timeout option and activate tcp keepalive.

Here is the stack where the program hang:
![image](https://user-images.githubusercontent.com/11644714/39693063-5274b014-5215-11e8-8041-e9822137ad1b.png)

My redis servers run in cluster mode by using twem proxy, when this problem occurred, there is one redis server shutdown due to hardware problem, and the MGet command right just sent to this broken redis server.
Unlike the other hang problems reported, my client program has set timeout and keepalive option when connected to redis servers, and although the client program runs in multi-thread task, but would use one exclusive hiredis client context for each thread. So would that be some new problems?

Gwan-He:
Thanks for reply. I have read the hiredis code and I think the problem might caused by reconnect.
When my client program first connects to proxy, client calls redisSetTimeout and redisEnableKeepAlive api to set timeout and enable tcp keep alive. But after some errors occurred, the client would call redisReconnect api to reconnect to proxy, and the problem just happened here: redisReconnect api wouldn't set timeout and keep alive.
As described by the redisReconnect api comment, the reconnect api would rebuilt the connection with exactly the same config as old one, but seems this api doesn't implemented as the comment described. So is this a bug or the hiredis just implement in this way?

hippo-dalaoshe:
have you solve this problem? I meet the same problem when I reset the IPSEC between my hiredis client and redis server, my client program hang until the IPSEC reconnect and the hiredis context->err is the error "REDIS_ERR_IO", I want to find this disconnect as soon as possible instead of when the IPSEC reconnect

catterer:
I believe I reproduced the same issue. It seems that after call to redisReconnect you need to re-set timeout using redisSetTimeout, otherwise the client hangs in redisBufferRead. I reproduced this on current master (2d9d775) as well as older version (0.13.3).

Here is what I did:
I listen for a connection but never accept it. For example, you can do it with this script:
```
#!/usr/bin/perl -w

use strict;
use Socket;

my $port = 5198;
my $proto = getprotobyname('tcp');

[...]

Issue #517, opened 2017-04-10
Hiredis Hang on redisBufferRead


naugaria7:
Hi All,

I have connection pool from which Redis Contexts are shared across different threads. HiRedis Hangs on RedisBufferRead() with following bt:

#0  0x0000000808bd6d0c in read () from /lib/libc.so.7

#1  0x000000080859a830 in read () from /lib/libthr.so.3
#2  0x0000000803c01b07 in redisBufferRead () from ./lib/libhiredis.so.0.11
#3  0x0000000803c03006 in redisGetReply () from ./lib/libhiredis.so.0.11
#4  0x0000000803c03084 in redisvCommand () from ./lib/libhiredis.so.0.11
#5  0x0000000806d99486 in RedisUtil::run_redis_query_set (command=@0x7ffff73b5350) at RedisUtil.cpp:60
#6  0x0000000806d9e422 in RedisUtil::set_redis_session (sess_id=@0x7ffff73b5d60, scope_map_temp=<value optimized out>, is_external=false) at RedisUtil.cpp:122

I see in other posts that this could happen if same Context is used by two threads at same time. Since I am using a pool, this should not have caused the issue. Any leads would be appreciated.

Chinmay

michael-grunder:
Nothing actionable here and I'm guessing it's been resolved since this is from 2017

Issue #263, opened 2014-08-13
hiredis hangs on redisBufferRead


saravanan-elangovan:
We use hiredis to connect to redis server in block mode. Our application runs in RHEL 4 32 bit. It works fine and after some time while trying to do an operation, it hangs. On analysis, found it got hanged in the redisBufferRead. This application act like a switch which gets the data from TCP Socket and store the same in the Redis database for recoverability.

On netstat output, We found the SendQ of the our client has non-zero value which indicates that  request has not reached redis server. This behaviour happens with INCR & INCRBY mostly. 

We opened a redis-client and did the same operation and it works fine. 

gdb output
#0  0x008787a2 in _dl_sysinfo_int80 () from /lib/ld-linux.so.2
#1  0x001de69b in __read_nocancel () from /lib/tls/libpthread.so.0
#2  0x08073a42 in redisBufferRead (c=0x89c2fb8) at hiredis.c:1142
#3  0x08073c59 in redisGetReply (c=0x89c2fb8, reply=0x590dcbc) at hiredis.c:1228
#4  0x08073e99 in redisvCommand (c=0x89c2fb8, format=0x8077f40 "INCRBY %s %ld", ap=0x590dcf8 "t\021\225\b\001") at hiredis.c:1332
#5  0x08073ed2 in redisCommand (c=0x89c2fb8, format=0x8077f40 "INCRBY %s %ld") at hiredis.c:1349

shoutrain:
I met the same issue in 2012 and 2013.  And in these days I met them again.  Maybe it's due to the old lib, but why this issue is still keeping open status.

enviroment:
===============================
1. Linux tesla001 2.6.32-504.12.2.el6.x86_64 #1 SMP Wed Mar 11 22:03:14 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux. 
2. libhiredis.so.0.10 (yum updated from default centos 6 repo)

thread stacks:
===============================
[Switching to thread 3 (Thread 0x7fdbee57b700 (LWP 7795))]#0  0x000000348c40e75d in read () fro                                                                                                      m /lib64/libpthread.so.0
(gdb) bt
#0  0x000000348c40e75d in read () from /lib64/libpthread.so.0
#1  0x00007fdbeeddd7bb in redisBufferRead () from /usr/lib64/libhiredis.so.0.10
#2  0x00007fdbeeddefcf in redisGetReply () from /usr/lib64/libhiredis.so.0.10
[...]

PR #986, opened 2021-10-03
Fix hiredis blocking indefinitely despite timeout being set (#593)

catterer:
Currently, `redisSetTimeout` function does not update `command_timeout` field in the context.
The following call to `redisReconnect` then resets the socket timeout to infinity.
This is probably the root cause of the issues (#593), (#718), (#263), (#517).

This PR fixes the bug and also fixes the corresponding test, which had two issues:
1. DEBUG SLEEP 3 was sent twice because manual `c->funcs->write` did not flush the connection buffer
2. DEBUG SLEEP 3 in one test affected the following test

### DIFF
--- a/test.c
+++ b/test.c
@@ -1107,7 +1107,13 @@ static void test_blocking_connection_timeouts(struct config config) {
     test("Does not return a reply when the command times out: ");
     if (detect_debug_sleep(c)) {
         redisAppendFormattedCommand(c, sleep_cmd, strlen(sleep_cmd));
+
+        // flush connection buffer without waiting for the reply
         s = c->funcs->write(c);
+        assert(s == (ssize_t)sdslen(c->obuf));
+        sdsfree(c->obuf);
+        c->obuf = sdsempty();
+
         tv.tv_sec = 0;
         tv.tv_usec = 10000;
         redisSetTimeout(c, tv);
@@ -1120,6 +1126,9 @@ static void test_blocking_connection_timeouts(struct config config) {
                   strcmp(c->errstr, "recv timeout") == 0);
 #endif
         freeReplyObject(reply);
+
+        // wait for the DEBUG SLEEP to complete so that Redis server is unblocked for the following tests
+        sleep(3);
     } else {
         test_skipped();
     }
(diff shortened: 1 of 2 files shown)


test.c
```diff
@@ -1107,7 +1107,13 @@ static void test_blocking_connection_timeouts(struct config config) {
     test("Does not return a reply when the command times out: ");
     if (detect_debug_sleep(c)) {
         redisAppendFormattedCommand(c, sleep_cmd, strlen(sleep_cmd));
+
+        // flush connection buffer without waiting for the reply
```
bjosv, line 1111: An alternative could be to use `redisBufferWrite()` instead, which resets the `obuf`

test.c
```diff
@@ -1120,6 +1126,9 @@ static void test_blocking_connection_timeouts(struct config config) {
                   strcmp(c->errstr, "recv timeout") == 0);
 #endif
         freeReplyObject(reply);
+
+        // wait for the DEBUG SLEEP to complete so that Redis server is unblocked for the following tests
+        sleep(3);
     } else {
         test_skipped();
     }
```
bjosv, line 1134: It could be good to add another testcase after this which covers the found fault.
This would do something like:
```
do_reconnect(c, config); /* Should set a command_timeout */
redisAppendFormattedCommand(c, sleep_cmd, strlen(sleep_cmd));
s = c->funcs->write(c);
..
reply = redisCommand(c, "GET foo");
test_cond(s > 0 && reply == NULL && c->err == REDIS_ERR_IO && strcmp(c->errstr, "Resource temporarily unavailable") == 0);
```
Maybe the `detect_debug_sleep()` could cache its first result to avoid communication problems when/if used in the new testcase.

bjosv (commented):
Nice finding!

michael-grunder:
Thanks for the PR.

I'm in favor of merging this, but I'm wondering whether this is _technically_ a breaking change?

[Relevant XKCD](https://imgs.xkcd.com/comics/workflow.png)

michael-grunder:
Closing in favor of #1093 (which is just this PR rebased against master).

PR #989, opened 2021-10-05
Fix cmake config path on Linux.

xkszltl:
CMake config files were installed to `/usr/local/share/hiredis`, which is not recognizable by `find_package()`.
I'm not sure why it was set that way.
Given the commit introducing it is for Windows, I keep that behavior consistent there, but fix the rest.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,7 +123,11 @@ export(EXPORT hiredis-targets
     FILE "${CMAKE_CURRENT_BINARY_DIR}/hiredis-targets.cmake"
     NAMESPACE hiredis::)
 
-SET(CMAKE_CONF_INSTALL_DIR share/hiredis)
+if(WIN32)
+    SET(CMAKE_CONF_INSTALL_DIR share/hiredis)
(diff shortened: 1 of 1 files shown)


michael-grunder:
I'm pretty unfamiliar with CMake but this does appear correct.  I'll read up a bit on CMake's `find_package` and then get it merged.

xkszltl:
Emmm....Actually after double checking I might be wrong. `/usr/local/share/hiredis` could be supported as well. Originally I saw it after packaging into rpm and think that doesn't look right, probably because rarely anyone uses that.

image

michael-grunder:
So can this be closed then?

xkszltl:
Up to you, but I would recommend to merge.
This PR fixed a "package not found" issue, although I'm not sure about the root cause because either way should be OK based on the doc.

chayim (approved):
Looks good - and even tested in a Windows VM and Linux docker.

michael-grunder:
Merged, thanks!

@chayim Thanks for the help!

PR #997, opened 2021-10-11
Makefile: move SSL options into a block and refine rules

pizhenwei:
Makefile: move SSL options into a block and refine rules
    
    Move SSL options into a block to make it easy to read, and change
    global rules to make it easy to maintain. For the further step, it
    gets extensible to add another type.
    
    New version of the library building rule:
            static: $(STLIBNAME) $(SSL_STLIB)
            dynamic: $(DYLIBNAME) $(SSL_DYLIB)
    
    Compare with the orignal version:
            dynamic: $(DYLIBNAME)
            static: $(STLIBNAME)
            ifeq ($(USE_SSL),1)
            dynamic: $(SSL_DYLIBNAME)
            static: $(SSL_STLIBNAME)
            endif
    
    If we want to add a new type(Ex, RDMA), for the new version, we can
    do like this:
            ########### RDMA options start ###############
            .....
            ########### RDMA options end   ###############
    
            static: $(STLIBNAME) $(SSL_STLIB) $(RDMA_STLIB)
            dynamic: $(DYLIBNAME) $(SSL_DYLIB) $(RDMA_DYLIB)
    
    Signed-off-by: zhenwei pi <pizhenwei@bytedance.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,16 +4,10 @@
 # This file is released under the BSD license, see the COPYING file
 
 OBJ=alloc.o net.o hiredis.o sds.o async.o read.o sockcompat.o
-SSL_OBJ=ssl.o
 EXAMPLES=hiredis-example hiredis-example-libevent hiredis-example-libev hiredis-example-glib hiredis-example-push
-ifeq ($(USE_SSL),1)
-EXAMPLES+=hiredis-example-ssl hiredis-example-libevent-ssl
-endif
 TESTS=hiredis-test
 LIBNAME=libhiredis
 PKGCONFNAME=hiredis.pc
-SSL_LIBNAME=libhiredis_ssl
-SSL_PKGCONFNAME=hiredis_ssl.pc
 
 HIREDIS_MAJOR=$(shell grep HIREDIS_MAJOR hiredis.h | awk '{print $$3}')
 HIREDIS_MINOR=$(shell grep HIREDIS_MINOR hiredis.h | awk '{print $$3}')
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -60,22 +55,38 @@ DYLIB_MAKE_CMD=$(CC) -shared -Wl,-soname,$(DYLIB_MINOR_NAME)
 STLIBNAME=$(LIBNAME).$(STLIBSUFFIX)
 STLIB_MAKE_CMD=$(AR) rcs
 
+uname_S := $(shell sh -c 'uname -s 2>/dev/null || echo not')
+
+# Deps (use make dep to generate this)
+alloc.o: alloc.c fmacros.h alloc.h
```
michael-grunder, line 61: I think this will need to be moved below the `all:` target since right now running make without args only builds `alloc.o`

michael-grunder:
Hi, thanks for the PR.

If you rebase your branch against master (and move the deps targets) I will test and get this merged.

pizhenwei:
> Hi, thanks for the PR.
> 
> If you rebase your branch against master (and move the deps targets) I will test and get this merged.

Force updated, thanks for your suggestion.

michael-grunder:
Looks like this breaks the tests in Windows, because it's unconditionally adding `-lssl -lcrypto -lpthread` regardless of the `USE_SSL` value.

This [small modification on my fork](https://github.com/michael-grunder/hiredis/commit/89f5c454ac5748319118ac4243ea7766800fe649) seems to solve the problem.

If that looks OK to you I can include the patch when I merge.

pizhenwei:
> Looks like this breaks the tests in Windows, because it's unconditionally adding `-lssl -lcrypto -lpthread` regardless of the `USE_SSL` value.
> 
> This [small modification on my fork](https://github.com/michael-grunder/hiredis/commit/89f5c454ac5748319118ac4243ea7766800fe649) seems to solve the problem.
> 
> If that looks OK to you I can include the patch when I merge.

Thanks!

michael-grunder:
Merged, thanks!

PR #1010, opened 2021-10-25
Add asynchronous API tests

bjosv:
Add asynchronous tests for pubsub (RESP2)
    
This testcase verifies the asynchronous handling of the pubsub flow.
The testcase subscribes to a channel while a second client is used to publish a message on the channel.
After receiving the published message the testcase will unsubscribe and disconnect.
  
This testcase can be modified to reproduce an issue with having an incoming published message while unsubscribing to the channel.


bjosv:
@michael-grunder Is this inline with your thoughts regarding async testing?

michael-grunder:
@bjosv Yes, they're great.

Do you have any issues with my minor tweaks in 7ad38dc4a87393efeca405df8b14a83eb43158b9?  Without the changes I run into some compilation issues building on a system with both libev and libevent.

I didn't look in detail but it [seems related to this issue](https://github.com/libevent/libevent/issues/596)

bjosv:
@michael-grunder Your change looks fine to me.

I have both libev and libevent installed, but it seems that my installed libev-dev apt-package on Ubuntu 20.04 don't contain the file event.h. When I now check libevents file `event.h` it states that `<event2/event.h>` should be used, so your change is the right way to do it for many reasons!

michael-grunder:
Merged, thanks!

PR #1011, opened 2021-10-27
Trigger CI failure when valgrind issues are found

bjosv:
By default Valgrind will return the exit code from the tested process.
Since our test can return 0 (`ALL TESTS PASS`) even when a leak was found we need to tell Valgrind to return an error code instead. This will trigger a CI-job failure when issues are found.

See `--error-exitcode` in [man](https://linux.die.net/man/1/valgrind) for details.


bjosv:
@michael-grunder This might be a good one to merge, the exitcode can be changed if another non-nil value is preferred.

PR #1012, opened 2021-10-28
Add asynchronous test for pubsub using RESP3

### LINKED ISSUE #967: Why does hiredis ignore my callback on UNSUBSCRIBE in RESP3?
When I `SUBSCRIBE` to a channel with a callback, `hiredis` calls it with `subscribe` and `message` push messages, but not with `unsubscribe`. Instead, `hiredis` would send the reply to a push handler. This makes it harder for me to track callback lifetimes in C++, since I no longer have access to `void *privdata` that I have attached to my subscription.

When running the following modified `example-libuv` (and manually publishing a message after subscribing), I was expecting the following output:
```
subscribed
message
unsubscribed
```
Instead, I get:
```
subscribed
message
push: unsubscribe
```

```c
#include <adapters/libuv.h>
#include <async.h>
#include <hiredis.h>

void callback(redisAsyncContext *c, void *r, void *privdata) {
  redisReply *reply = r;
  char *what = reply->element[0]->str;

  if (strcmp(what, "subscribe") == 0) {
    printf("subscribed\n");
    return;
  }

  if (strcmp(what, "message") == 0) {
    printf("message\n");
    redisAsyncCommand(c, NULL, NULL, "UNSUBSCRIBE");
    return;
  }

  if (strcmp(what, "unsubscribe") == 0) {
    printf("unsubscribed\n");
    redisAsyncFree(c);
  }
}

void pushCallback(redisAsyncContext *c, void *r) {
  redisReply *reply = r;
  char *what = reply->element[0]->str;
  printf("push: %s\n", what);
  redisAsyncFree(c);
}

int main(int argc, char **argv) {

  if (c->err) {
    return 1;
  }

  uv_loop_t *loop = uv_default_loop();
  redisLibuvAttach(c, loop);

  redisAsyncSetPushCallback(c, pushCallback);
  redisAsyncCommand(c, NULL, NULL, "HELLO 3");
  redisAsyncCommand(c, callback, NULL, "SUBSCRIBE foo");

  uv_run(loop, UV_RUN_DEFAULT);
  return 0;
}
```

It appears that this can be fixed with a one line patch:
```
--- a/async.c
+++ b/async.c
@@ -496,7 +496,8 @@ static int redisIsSubscribeReply(redisReply *reply) {
     len = reply->element[0]->len - off;
 
     return !strncasecmp(str, "subscribe", len) ||
-           !strncasecmp(str, "message", len);
+           !strncasecmp(str, "message", len) ||
+           !strncasecmp(str, "unsubscribe", len);
 
 }
 ```
 Modified `hiredis` passes all unit tests and produces the output I expected.

bjosv:
This PR adds a testcase for pubsub using RESP3 and reuses the subscribe callback from the RESP2 testcase,
all to make sure we have a common behavior.
This requires a correction for `unsubscribe` responses which previously was given to the push callback.

Fixes #967

### DIFF
--- a/async.c
+++ b/async.c
@@ -496,8 +496,8 @@ static int redisIsSubscribeReply(redisReply *reply) {
     len = reply->element[0]->len - off;
 
     return !strncasecmp(str, "subscribe", len) ||
-           !strncasecmp(str, "message", len);
-
+           !strncasecmp(str, "message", len) ||
+           !strncasecmp(str, "unsubscribe", len);
 }
 
 void redisProcessCallbacks(redisAsyncContext *ac) {
--- a/test.c
+++ b/test.c
@@ -1453,6 +1453,7 @@ struct event_base *base;
 typedef struct TestState {
     redisOptions *options;
     int           checkpoint;
+    int           resp3;
 } TestState;
 
 /* Testcase timeout, will trigger a failure */
@@ -1479,7 +1480,7 @@ void publish_msg(redisOptions *options, const char* channel, const char* msg) {
     disconnect(c, 0);
 }
 
-/* Subscribe callback for test_pubsub_handling:
+/* Subscribe callback for test_pubsub_handling and test_pubsub_handling_resp3:
  * - a published message triggers an unsubscribe
  * - an unsubscribe response triggers a disconnect */
 void subscribe_cb(redisAsyncContext *ac, void *r, void *privdata) {
@@ -1487,7 +1488,7 @@ void subscribe_cb(redisAsyncContext *ac, void *r, void *privdata) {
     TestState *state = privdata;
 
     assert(reply != NULL &&
-           reply->type == REDIS_REPLY_ARRAY &&
+           reply->type == (state->resp3 ? REDIS_REPLY_PUSH : REDIS_REPLY_ARRAY) &&
            reply->elements == 3);
 
     if (strcmp(reply->element[0]->str,"subscribe") == 0) {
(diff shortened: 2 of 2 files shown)


michael-grunder:
Thanks, @bjosv I will go through this and get it merged this week.

bjosv:
Now rebased and updated to match the existing asynchronous testing on master.

bjosv:
@michael-grunder Do you think handling the `(p)unsubscribe` in the subscribe-callback for RESP3 would be problematic due to backward compatibility?

michael-grunder:
@bjosv I'll take a closer look, although I'm thinking v2.0.0 is probably good to release soon so we can fix [this hack](https://github.com/redis/hiredis/tree/unique-sds-symbols) of the sds symbols we needed to do for Redis.

zuiderkwast:
Why 2.0.0 and not just 1.1.0? Are there any backwards-incompatible changes needed to remove the sds hack?

There are as well differences in dict between Redis and Hiredis and probably more changes to come in Redis.

michael-grunder:
@zuiderkwast We may be able to get away with `v1.1.0` which would be fine with me.  The issue, I suppose is that we are technically exporting the sds symbols so I don't think we can change their names (at least not by default) without a major version bump.

It's been a while since I looked at that branch though, we may be able to `#ifndef` our way around the conflict.

For `v2.0.0` it's probably prudent to hide the internals of `redisContext` behind an opaque struct.  Over the years this has caused many issues, especially for people new to C.

PR #1013, opened 2021-10-29
BSD linkage fix proposal

devnexen:


michael-grunder:
Should we limit this extra linker argument only to FreeBSD.  We don't need it in Linux.

I can also take a look at the Makefile proper

rajhlinux:
how to build on freebsd 13.1?

michael-grunder:
I don't have a BSD machine so I'm not sure.  Are you encountering a specific error?

Issue #1015, opened 2021-11-03
Why no `scheduleTimer` in libuv adapter?


MichaelSuen-thePointer:
https://github.com/redis/hiredis/blob/master/adapters/libuv.h#L100

Is it a missing feature or it is because libuv doesn't support it?

MichaelSuen-thePointer:
Anyway, I added it myself, will make a PR soon

PR #1016, opened 2021-11-03
Add timeout support for libuv adapter

MichaelSuen-thePointer:
see #1015 

sorry for no test cases : (


michael-grunder:
Thanks, looks good on first glance.  I know it's somewhat trivial but would you mind re-indenting with 4 spaces?

Then I'll get it merged.

MichaelSuen-thePointer:
@michael-grunder 
done :)

zuiderkwast (approved):
Indentation looks fine. Didn't review the actual code. :see_no_evil:

michael-grunder:
Hi, this code asserts when testing against the `0xDEADBEEF` sentinel value.

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <signal.h>

#include <hiredis.h>
#include <async.h>
#include <adapters/libuv.h>
#include <time.h>
#include <sys/time.h>

void sleepCb(redisAsyncContext *c, void *r, void *privdata) {
    printf("sleep CB\n");
}

#ifndef _WIN32
    signal(SIGPIPE, SIG_IGN);
#endif

    uv_loop_t* loop = uv_default_loop();

    if (c->err) {
        /* Let *c leak for now... */
        return 1;
    }

    redisLibuvAttach(c,loop);
    redisAsyncSetTimeout(c, (struct timeval){ .tv_sec = 1, .tv_usec = 0});
    redisAsyncCommand(c, sleepCb, NULL, "DEBUG SLEEP %f", 1.5);
    uv_run(loop, UV_RUN_DEFAULT);
    return 0;
}```

```bash
❯ ./uvtm
sleep CB
uvtm: /home/mike/dev/hiredis/adapters/libuv.h:83: on_handle_close: Assertion `p->timer.data == (void *)0xDEADBEEF' failed.
```

I'm wondering why we need the sentinel value at all

MichaelSuen-thePointer:
@michael-grunder 
I was assuming the timer destroy callback to be earlier than the socket destroy callback, seems I was wrong. I will look into it soon.

> edit: I took a shot at adding a timeout to the libuv adapter and came up with this.

I don't know if it is ok to just stop the timer and not 'destroy' it, but it should be good.

MichaelSuen-thePointer:
@michael-grunder 
Ok fixed, now I use private data part of the uv handle as a guard to determine whether to call `hi_free(p)` or not. Now your example passed, and I added it to the example directory.

BTW, It is necessary to call `uv_close` on all of its handle type.

Makefile
```diff
@@ -166,6 +166,8 @@ hiredis-example-macosx: examples/example-macosx.c adapters/macosx.h $(STLIBNAME)
 hiredis-example-ssl: examples/example-ssl.c $(STLIBNAME) $(SSL_STLIBNAME)
 	$(CC) -o examples/$@ $(REAL_CFLAGS) -I. $< $(STLIBNAME) $(SSL_STLIBNAME) $(REAL_LDFLAGS) $(SSL_LDFLAGS)
 
+hiredis-example-uvtm: examples/example-uvtm.c adapters/libuv.h $(STLIBNAME) $(SSL_STLIBNAME)
```
bjosv, line 169: We already have an example for libuv some rows below. It might be better to put this new rule close to the existing one?
A suggestion is to rename this new example to something like `hiredis-example-libuv-timeout` to match the existing naming.
Maybe the existing one can be used for the timeout case?

examples/example-uvtm.c
```diff
@@ -0,0 +1,42 @@
...
+#include <signal.h>
+
+#include <hiredis.h>
+#include <async.h>
+#include <adapters/libuv.h>
+#include <time.h>
+#include <sys/time.h>
+
+void sleepCb(redisAsyncContext *c, void *r, void *privdata) {
+    printf("sleep CB\n");
```
bjosv, line 20: When the async command timeout the reply should be `NULL`. This can be shown in this example by using
```
    redisReply *reply = r;
    assert(reply == NULL);
```
or by some simple printf.

examples/example-uvtm.c
```diff
@@ -0,0 +1,42 @@
+
```
bjosv, line 1: We have an existing `example-libuv.c`, rename this to something like `example-libuv-timeout.c`?

Makefile
bjosv, line 169: Add the example to the `CMakeLists.txt` in `examples/` folder as well.
See how `example-libuv` is added and add a `ADD_EXECUTABLE` and `TARGET_LINK_LIBRARIES` to the new example.

examples/example-uvtm.c
```diff
@@ -0,0 +1,42 @@
...
+#include <async.h>
+#include <adapters/libuv.h>
+#include <time.h>
+#include <sys/time.h>
+
+void sleepCb(redisAsyncContext *c, void *r, void *privdata) {
+    printf("sleep CB\n");
+}
+
+int main (int argc, char **argv) {
```
bjosv, line 23: warning: unused parameter `argc` and `argv`

examples/example-uvtm.c
```diff
@@ -0,0 +1,42 @@
...
+#include <string.h>
+#include <signal.h>
+
+#include <hiredis.h>
+#include <async.h>
+#include <adapters/libuv.h>
+#include <time.h>
+#include <sys/time.h>
+
+void sleepCb(redisAsyncContext *c, void *r, void *privdata) {
```
bjosv, line 19: warning: unused parameter `c`, `r`, `privdata`

bjosv (commented):
Nice, I have some suggestions.

MichaelSuen-thePointer:
push -f to fix commit message, sorry
@bjosv I just merged example-uvtm with example-libuv and removed example-uvtm

adapters/libuv.h
```diff
@@ -7,111 +7,154 @@
...
+    if (!p->timer.data) {
+        // timer is uninitialized
+        if (uv_timer_init(p->handle.loop, &p->timer) != 0) {
+            return;
+        }
+        p->timer.data = p;
+    }
+    // updates the timeout if the timer has already started
+    // or start the timer
+    uv_timer_start(&p->timer, (uv_timer_cb)redisLibuvTimeout, millsec, 0);
```
michael-grunder, line 111: Casting a function pointer in this way is undefined behavior and may crash in weird and hard to track down ways.

I'm not overly familiar with libuv but I'm guessing there is a way to `#ifdef` around it?

MichaelSuen-thePointer: Done, it turns out I was using a very old libuv version (0.10.x), however I added the `#ifdef` and it is ok now.

adapters/libuv.h
```diff
@@ -90,7 +90,14 @@ static void on_handle_close(uv_handle_t *handle) {
     // else, wait for `on_timer_close`
 }
 
+#if UV_MAJOR_VERSION > 0 || UV_MINOR_VERSION > 11 || (defined(UV_VERSION_PATCH) && UV_VERSION_PATCH > 22)
```
michael-grunder, line 93: How about this instead:

```suggestion
#if !defined(UV_VERSION_PATCH) || (UV_MAJOR_VERSION < 1 && UV_MINOR_VERSION < 12 && UV_VERSION_PATCH < 23)
```

Looks like `UV_VERSION_PATCH` was introduced in libuv 0.10.2 so if it's not defined, we want the old prototype.
libuv/libuv@c7b1c53ef1afe7c0fef0822d72b0783368b626de

MichaelSuen-thePointer: looks cleaner👍

adapters/libuv.h
```diff
@@ -90,7 +90,7 @@ static void on_handle_close(uv_handle_t *handle) {
     // else, wait for `on_timer_close`
 }
 
-#if UV_MAJOR_VERSION > 0 || UV_MINOR_VERSION > 11 || (defined(UV_VERSION_PATCH) && UV_VERSION_PATCH > 22)
+#if !defined(UV_VERSION_PATCH) || (UV_MAJOR_VERSION < 1 && UV_MINOR_VERSION < 12 && UV_VERSION_PATCH < 23)
 // libuv removed `status` parameter since v0.11.23
```
zuiderkwast, line 94: This is maybe wrong for versions like 0.10.37 (it exists).

minor < 12 doesn't always mean minor == 11.

MichaelSuen-thePointer: Oh, thanks for the note, you are correct, `major.minor.patch < 0.11.23` expands to something like `major < 0 || (major == 0 && minor < 11) || (minor == 11 && patch < 23)`

MichaelSuen-thePointer: Let me try reduce it, could you lend some help?

zuiderkwast: Major < 0 can't happen so this part can be removed.

I think `major == 0 && (minor < 11 || (minor == 11 && patch < 23))`

MichaelSuen-thePointer:
Fixed a silly typo, corrected version check macro, thanks @zuiderkwast for the help

adapters/libuv.h
```diff
@@ -90,13 +90,21 @@ static void on_handle_close(uv_handle_t *handle) {
...
-#else
+// explain:
+// 1. !defined(UV_VERSION_PATCH), must < 0.10.2
+// 2. major.minor.patch < 0.11.23
+// => (major < 0) || (major == 0 && minor < 11) || (minor == 11 && patch < 23)
+//    major < 0 can't happen
+// => (major == 0 && minor < 11) || (minor == 11 && patch < 23)
+
+//          < 0.10.2 goes here                  < 0.10.2~x goes here                                     0.11.0~22 goes here
+#if !defined(UV_VERSION_PATCH) || (UV_VERSION_MAJOR == 0 && UV_VERSION_MINOR < 11) || (UV_VERSION_MINOR == 11 && UV_VERSION_PATCH < 23)
```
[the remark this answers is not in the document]
zuiderkwast: I think the last example with linebreaks is the clearest. It doesn't even require comments to understand it. (Maybe only for the part !defined patch.)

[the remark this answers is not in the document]
zuiderkwast: Actually, it will work even if patch is not defined, because `(UV_VERSION_MAJOR == 0 && UV_VERSION_MINOR < 11)` will match for 0.10.x before patch is checked.

So I think we only need this:

```suggestion
#if (UV_VERSION_MAJOR == 0 && UV_VERSION_MINOR < 11) || \
    (UV_VERSION_MAJOR == 0 && UV_VERSION_MINOR == 11 && UV_VERSION_PATCH < 23)
```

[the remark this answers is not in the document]
MichaelSuen-thePointer: OK, sure😂

adapters/libuv.h
```diff
@@ -100,7 +100,8 @@ static void on_handle_close(uv_handle_t *handle) {
 // => (major == 0 && minor < 11) || (minor == 11 && patch < 23)
 
 //          < 0.10.2 goes here                  < 0.10.2~x goes here                                     0.11.0~22 goes here
```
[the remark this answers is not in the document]
MichaelSuen-thePointer: done

michael-grunder:
Merged, thanks!

PR #1019, opened 2021-11-18
fix the case of errno=11 and BLOCKING read mode where we should retry

subhasisb:
If the client is in BLOCKING mode, then EAGAIN should be a transient problem and should not result in error. I believe there is a single character mistake in the code block where EAGAIN is checked, the ! should not have been there.

### DIFF
--- a/net.c
+++ b/net.c
@@ -60,7 +60,7 @@ void redisNetClose(redisContext *c) {
 ssize_t redisNetRead(redisContext *c, char *buf, size_t bufcap) {
     ssize_t nread = recv(c->fd, buf, bufcap, 0);
     if (nread == -1) {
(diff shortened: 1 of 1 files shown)


michael-grunder:
Well it _certainly_ shouldn't be an error in non-blocking mode, and this would make it one, no?

subhasisb:
> Well it _certainly_ shouldn't be an error in non-blocking mode, and this would make it one, no?

I am not sure whether we should or not error in case of non-blocking mode. We can, of course, modify it so that it is never an error and is always retried. However, one could argue that in case of non-blocking mode, the client code would want to retry the operation again, instead of the API itself retrying?

subhasisb:
Withdrawing this PR as I am making changes to the caller itself to avoid the situation

PR #1027, opened 2021-12-10
Support PING while subscribing (RESP2)

### LINKED ISSUE #351: PING not allowed while subscribing
This is in reference to antirez/redis#420, specifically https://github.com/antirez/redis/issues/420#issuecomment-128437282 

The reason hiredis silently ignores the `ping` response is because it stores `subscribe` and `psubscribe` callbacks in separate dictionaries in order to deliver messages to the correct handler.

It looks for a redis publish message, which can take the following forms (with `pong` being new):

| SUBSCRIBE | PSUBSCRIBE | PONG |
| --- | --- | --- |
| 1) "message" | 1) pmessage | 1) pong |
| 2) "foo" | 2) "foo*" | 2) pongarg |
| 3) "bar" | 3) "foo" |  |
|  | 4) "bar" |  |

Hiredis is only accounting for `message` or `pmessage`, and specifically just checks the first character of the first multibulk reply string to determine which one it is, which is done [here](https://github.com/redis/hiredis/blob/master/async.c#L378).  Given that `pong` starts with `p` it looks in the `psubscribe` dict.

It uses the second argument (which for subscribe or psubscribe will be the channel/pattern) to search, but in the case of `pong` can now be literally anything.

You can actually "trick" hiredis into executing your callback by doing something like this:

``` c
    signal(SIGPIPE, SIG_IGN);

    redisLibevAttach(EV_DEFAULT_ c);
    redisAsyncSetConnectCallback(c,connectCallback);
    redisAsyncSetDisconnectCallback(c,disconnectCallback);

    redisAsyncCommand(c, subscribeCallback, NULL, "PSUBSCRIBE pong*");

    ev_loop(EV_DEFAULT_ 0);
    return 0;
}
```

``` c
void subscribeCallback(redisAsyncContext *c, void *r, void *privdata) {
    printf("In subscribeCallback...\n");

    redisReply *reply = r;
    int i;

    if (reply == NULL) return;

    for (i = 0; i < reply->elements; i++) {
        if (reply->element[i]->type == REDIS_REPLY_STRING) {
            printf("%i: %s\n", i, reply->element[i]->str);
        }
    }
    redisAsyncCommand(c, subscribeCallback, NULL, "PING pong*");
}
```

This is obviously not a solution of any kind but is what's going on.  Hiredis could be made to handle the new message types by detecting "pong" specifically, and using the argument as either the channel name or pattern.  This would mean that you couldn't add any other kind of context to the ping, but would allow you to use it as a mechanism to detect the connection was still live.

bjosv:
This PR tunes the matching of pushed message replies to allow handling of `PING`.
Since `PING`s response `PONG` consists of 2 elements its callback will now be found and called.

Since this command is valid to send during subscribing, and most commands in RESP3, the callback list is renamed from `invalid` to `replies` to detail this fact (same as in PR #1014).

Fixes #351

### DIFF
--- a/async.c
+++ b/async.c
@@ -419,10 +419,11 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
     char *stype;
     sds sname;
 
-    /* Custom reply functions are not supported for pub/sub. This will fail
-     * very hard when they are used... */
-    if (reply->type == REDIS_REPLY_ARRAY || reply->type == REDIS_REPLY_PUSH) {
-        assert(reply->elements >= 2);
+    /* Match reply with the expected format of a pushed message.
+     * The type and number of elements (3 to 4) are specified at:
+     * https://redis.io/topics/pubsub#format-of-pushed-messages */
+    if ((reply->type == REDIS_REPLY_ARRAY && reply->elements >= 3) ||
+        reply->type == REDIS_REPLY_PUSH) {
         assert(reply->element[0]->type == REDIS_REPLY_STRING);
         stype = reply->element[0]->str;
         pvariant = (tolower(stype[0]) == 'p') ? 1 : 0;
--- a/async.c
+++ b/async.c
@@ -466,7 +466,7 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
         }
         sdsfree(sname);
     } else {
-        /* Get callback for pending command during subscribe. */
+        /* Shift callback for pending command in subscribed context. */
         __redisShiftCallback(&ac->sub.replies,dstcb);
     }
     return REDIS_OK;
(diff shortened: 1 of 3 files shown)


async.c
```diff
@@ -467,8 +466,8 @@ static int __redisGetSubscribeCallback(redisAsyncContext *ac, redisReply *reply,
         }
         sdsfree(sname);
     } else {
-        /* Shift callback for invalid commands. */
-        __redisShiftCallback(&ac->sub.invalid,dstcb);
+        /* Get callback for pending command during subscribe. */
+        __redisShiftCallback(&ac->sub.replies,dstcb);
```
zuiderkwast, line 470: Why replace "Shift" with "Get" in this comment?

Shift means pop from the beginning of the list. This meaning of `shift` is known from shell scripting.

zuiderkwast (approved):
LGTM

michael-grunder:
Merged, thanks again!

PR #1030, opened 2021-12-16
Fix integer overflow when format command larger than 4GB

sundb:
See https://groups.google.com/g/redis-db/c/loPFJAqDIgI/m/GfmBKX4gDAAJ?utm_medium=email&utm_source=footer.

This PR mainly fixes two integer overflow.
1. When command arg is larger than 4GB, `hi_sdscatfmt` should use `%U`, not `%u`.
2. When the command is larger than 4GB, the return type of `redisFormatSdsCommandArgv` should be `long long`.


michael-grunder:
Merged, thanks!

PR #1034, opened 2021-12-28
Allow sending commands after sending an unsubscribe

### LINKED ISSUE #968: Incorrect handling of commands in pub/sub state with RESP3
`hiredis` misbehaves when you upgrade the connection to RESP3 and then try to run commands other than `SUBSCRIBE`/`UNSUBSCRIBE`.

Both Redis [source code](https://github.com/redis/redis/blob/f03af47a34ec672a7d9b18150a5be3a83681c19b/src/server.c#L4144) and [documentation](https://github.com/redis/redis-doc/blame/master/topics/client-side-caching.md#L117) seem to agree that it should be possible to run other commands. For example, the following should work:
```
HELLO 3
SUBSCRIBE foo
PUBLISH foo bar
PING
GET baz
UNSUBSCRIBE FOO
```

Unfortunately, when trying to run this, I trigger the following [assertion](https://github.com/redis/hiredis/blob/f5f31ff9b92b6bdf628716449d0d0782ceb7704a/async.c#L565):
```
/* No more regular callbacks and no errors, the context *must* be subscribed or monitoring. */
```
...because [here](https://github.com/redis/hiredis/blob/f5f31ff9b92b6bdf628716449d0d0782ceb7704a/async.c#L816), `hiredis` does not push user callbacks into `replies` when in pub/sub mode:
```c
if (c->flags & REDIS_SUBSCRIBED) {
    /* This will likely result in an error reply, but it needs to be
     * received and passed to the callback. */   // <- no longer true with RESP3
    if (__redisPushCallback(&ac->sub.invalid,&cb) != REDIS_OK)
        goto oom;
} else {
    if (__redisPushCallback(&ac->replies,&cb) != REDIS_OK)
        goto oom;
}
```

Additionally, [here](https://github.com/redis/hiredis/blob/f5f31ff9b92b6bdf628716449d0d0782ceb7704a/async.c#L541), it assumes that all regular callbacks are handled before entering pub/sub mode. I think this is also incorrect, since it should be possible to receive a reply to a command interleaved with subscription messages with RESP3.

I think that, in the async case, this can be fixed by:
- keeping track of current connection protocol in `redisContext`
- conditionally allowing other commands depending on protocol in `__redisAsyncCommand`
- reordering the code in `redisProcessCallbacks` to first handle all push messages, and then all others

I would like to help you with this, but I wanted to ask for some preliminary feedback on how a fix should look like.
I haven't tried the above using the synchronous API yet.

bjosv:
Instead of handling reply callbacks in two modes or states ("normal" state and "subscribed" state)
we now use a common holder for non-subscribe callbacks.
This fixes the current issue of putting the callback in the wrong holder when going from "subscribed" state to "regular" state.

Added a test to trigger the fault.

Fixes #968 (last part)

### DIFF
--- a/async.c
+++ b/async.c
@@ -144,8 +144,6 @@ static redisAsyncContext *redisAsyncInitialize(redisContext *c) {
 
     ac->replies.head = NULL;
     ac->replies.tail = NULL;
-    ac->sub.replies.head = NULL;
-    ac->sub.replies.tail = NULL;
     ac->sub.channels = channels;
     ac->sub.patterns = patterns;
 
@@ -312,8 +310,6 @@ static void __redisAsyncFree(redisAsyncContext *ac) {
     /* Execute pending callbacks with NULL reply. */
     while (__redisShiftCallback(&ac->replies,&cb) == REDIS_OK)
         __redisRunCallback(ac,&cb,NULL);
-    while (__redisShiftCallback(&ac->sub.replies,&cb) == REDIS_OK)
-        __redisRunCallback(ac,&cb,NULL);
 
     /* Run subscription callbacks with NULL reply */
     if (ac->sub.channels) {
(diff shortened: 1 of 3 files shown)


bjosv:
Found some issues with this solution, so I'm withdrawing this PR until fixed.
The testcase is still valid though.

PR #1038, opened 2022-01-11
Correction for command timeout during pubsub

bjosv:
A timeout of a non-subscribe command is ignored during pubsub,
and will be handled as an idle timeout and the response is still awaited for.

This correction triggers a disconnect instead, the same behavior as when not in the subscribed state.


bjosv:
@michael-grunder There is a lot of testcase in this PR, and can be seen as to much noise, so I'm open to just merging the correction if that's the case :)

PR #1040, opened 2022-01-21
Avoid incorrect call to the previous reply's callback

bjosv:
When multiple replies are parsed from a socket in one read;
a previously found callback might get reused when the current reply has no known callback.
    
This case is triggered in the added testcase `test_pubsub_multiple_channels()`,
which unsubscribes to the subscribed channels `A`, `B` and a non-subscribed channel `X`.
Without this correction a callback for wrong channel is called.
-  For the command `unsubscribe B X A`, `B`'s callback is called when handling the response for `X`.
-  Now this is not done, i.e. there is no callback called for `X`.

Monitoring was using this trait to repush its callback when all replies was parsed.
This is changed to repush for each message instead, which now also is tested with a new testcase.

### DIFF
--- a/async.c
+++ b/async.c
@@ -520,13 +520,6 @@ void redisProcessCallbacks(redisAsyncContext *ac) {
                 __redisAsyncDisconnect(ac);
                 return;
             }
-
-            /* If monitor mode, repush callback */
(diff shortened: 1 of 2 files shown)


michael-grunder:
Merged, thanks!

Issue #1031, opened 2021-12-21
Building on AIX fails with not expanding SIZE_MAX


scddev:
A fresh git clone fails to build on AIX.

Surprisingly alloc.c works while hiredis.c and all other files (read.c, sds.c, ...) which would use hi_*alloc methods fail not expanding SIZE_MAX.

```
~/hiredis $ cmake -S . -B build
-- The C compiler identification is GNU 6.3.0
-- The CXX compiler identification is GNU 6.3.0
-- Check for working C compiler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
Detected version: 1.0.3
-- Configuring done
-- Generating done
-- Build files have been written to: /home/scd/hiredis/build
[~/hiredis $ cmake --build build
gmake[1]: Entering directory '/home/scd/hiredis/build'
gmake[2]: Entering directory '/home/scd/hiredis/build'
Scanning dependencies of target hiredis
gmake[2]: Leaving directory '/home/scd/hiredis/build'
gmake[2]: Entering directory '/home/scd/hiredis/build'
[  5%] Building C object CMakeFiles/hiredis.dir/alloc.c.o
[ 10%] Building C object CMakeFiles/hiredis.dir/async.c.o
[ 15%] Building C object CMakeFiles/hiredis.dir/dict.c.o
[ 20%] Building C object CMakeFiles/hiredis.dir/hiredis.c.o
In file included from /home/scd/hiredis/hiredis.h:46:0,
                 from /home/scd/hiredis/hiredis.c:41:
/home/scd/hiredis/alloc.h: In function 'hi_calloc':
/home/scd/hiredis/alloc.h:64:9: error: 'SIZE_MAX' undeclared (first use in this function)
     if (SIZE_MAX / size < nmemb)
         ^~~~~~~~
/home/scd/hiredis/alloc.h:64:9: note: each undeclared identifier is reported only once for each function it appears in
[...]

PR #1043, opened 2022-01-24
fix building on AIX and SunOS (#1031)

scddev:
* handle AIX in `fmacros.h` to fix expansion
* link `socket` on Solaris to have missing symbols `send`, `recv`, ...


michael-grunder:
Merged, thank you so much!

PR #1045, opened 2022-02-01
Update hiredis sds with improvements found in redis

bjosv:
Update hiredis sds with improvements found in redis for convergence,
like the overflow mentioned in https://github.com/redis/redis/issues/9706

Handpicked from:
https://github.com/redis/redis/pull/4568
https://github.com/redis/redis/pull/9584
https://github.com/redis/redis/pull/8522

### DIFF
--- a/sds.c
+++ b/sds.c
@@ -90,7 +90,7 @@ sds sdsnewlen(const void *init, size_t initlen) {
     int hdrlen = sdsHdrSize(type);
     unsigned char *fp; /* flags pointer. */
 
-    assert(initlen + hdrlen + 1 > initlen); /* Catch size_t overflow */
+    if (hdrlen+initlen+1 <= initlen) return NULL; /* Catch size_t overflow */
     sh = s_malloc(hdrlen+initlen+1);
     if (sh == NULL) return NULL;
     if (!init)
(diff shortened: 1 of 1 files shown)


bjosv:
Ouch, it looks like CentOS8 has gone [EOL](https://www.centos.org/centos-linux-eol/) and that we need to do some updates for CI.

michael-grunder:
I actually ran into the CentOS issue yesterday [and even complained about it on Twitter](https://twitter.com/grumi78/status/1488338054100099074) :smile:

I believe there is a short-term solution to this.  I'll get that up now.

michael-grunder:
Thanks for porting these over.

I wonder if we can avoid the `assert`s on overflow though.  Simply return `NULL` instead so the error is just treated like an `OOM` for clients of the library.

bjosv:
> I wonder if we can avoid the `assert`s on overflow though. Simply return `NULL` instead so the error is just treated like an `OOM` for clients of the library.

Yes, you are right, much better. This follows how its handled in other places

bjosv:
Sorry, touchpad gone wild.. updated the PR with the change.

michael-grunder:
Merged, thanks!

PR #1048, opened 2022-02-15
cmake: set CMAKE_CONF_INSTALL_DIR to lib/cmake/hiredis.

cyfdecyf:
Installing cmake modules to `<PREFIX>/lib/cmake/` is more common and make it easy for users to use hiredis package.

If backwards compatibility should be kept, maybe we can install to both `share/hiredis` and `lib/cmake/hiredis`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ export(EXPORT hiredis-targets
     FILE "${CMAKE_CURRENT_BINARY_DIR}/hiredis-targets.cmake"
     NAMESPACE hiredis::)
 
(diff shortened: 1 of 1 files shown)


chayim:
@cyfdecyf In light of [this PR](https://github.com/redis/hiredis/pull/989) I think this needs revisiting. WDYT?

cyfdecyf:
I'll close this PR because `find_package(hiredis)` works with cmake configs files installing to `share/hiredis`.

I can't remember if I encounter any problem when I submit this PR. But seems the cmake doc about `find_package` search directory is right so `share/hiredis` can be used.

PR #1050, opened 2022-03-18
fix cmake version

smmir-cent:
Modify the CMake version for use in lower versions.


michael-grunder:
I was able to build hiredis in a container with CMake v3.0.0 so on the surface this change seems fine.

I'll go ahead and merge the changes, although I'm not really a CMake expert so we can revert if we run into issues in the future.

PR #1054, opened 2022-03-24
Embed debug information in windows static .lib file

kristjanvalur:
Embed debug information in windows static lib, rather than create a pdb fileþ

Using .pdb files with .lib files on windows is very inconvenient, particularly if the .lib
file is then linked as part of a different .dll.  Chances are that the original .pdb
will not be picked up or distributed along with the tooling.


michael-grunder:
Merged, thanks!

PR #1055, opened 2022-03-24
CMake: remove dict.c form hiredis_sources

Lipraxde:
Commit c6b8bd77c0fe00dbc455b39208f15761178160a3 to make all functions in dict.c static. If a CMake project set warning unused functions, and include hiredis using add_subdirectory , this cause warnings / errors.


bjosv (approved):
Nice to get CMake and make builds more alike, even that `dict` is not part of the external symbols (`nm -g`) nor included in the installed headers.
Building hiredis itself with something like `-Wall` today will also trigger the warnings you mentioned.

PR #1057, opened 2022-03-29
Use the same name for static and shared libraries

orgads:
On all system except MSVC, the targets are different.

Unix: libhiredis.so, libhiredis.a
MinGW: libhiredis.dll+libhiredis.dll.a, libhiredis.a
MSVC: hiredis.dll+hiredis.lib, hiredis_static.lib


kwach:
+1

orgads:
ping

michael-grunder:
Merged, thanks!

PR #1062, opened 2022-04-23
Handle push notifications before or after reply.

yossigo:
Current Redis unstable breaks tests, probably after redis/redis#9422. @michael-grunder I hope this fix makes sense, was attempting to maintain compatibility with both pre-fix and post-fix Redis versions.


michael-grunder:
Makes sense, thanks!

PR #1064, opened 2022-04-28
Fix warnings about implicit cast

TedLyngmo:
Added a few explicit casts:
unsigned char flags = (unsigned char)s[-1];

Signed-off-by: Ted Lyngmo <ted@lyncon.se>


TedLyngmo:
@michael-grunder  This is my first PR in a repo requiring workflow approval. Do I need to request this approval somehow?

bjosv:
> Any idea what could cause it to skip the tests?

You might already have found the reason, but the command `DEBUG SLEEP` which is needed by these testcases was introduced in Redis version 2.2.12. Some testcases requires an already running redis and there might be an old pre-2.2.12 version running on those distros by default.

TedLyngmo:
> > Any idea what could cause it to skip the tests?
> 
> You might already have found the reason, but the command `DEBUG SLEEP` which is needed by these testcases was introduced in Redis version 2.2.12. Some testcases requires an already running redis and there might be an old pre-2.2.12 version running on those distros by default.

Oh, ok, so there is currently no way to get patches through this gate - or what can I do?
I have some more ideas for patches to be able to enable more warnings. Will it be possible to get them in somehow?

michael-grunder:
I'm hesitant to merge changes to the `sds` library because this code is also in [Redis itself](https://github.com/redis/redis/blob/ff3a3577f29a07e965b3b38d07894410d28da499/src/sds.h#L88) so we try to minimise divergence.

How are you invoking the tests?  Treating skip as failures [is an option](https://github.com/redis/hiredis/blob/47c3ecefc04505dcc8e42b9f726fc29aa8084bec/test.c#L1932) that you don't have to set.

TedLyngmo:
> I'm hesitant to merge changes to the `sds` library because this code is also in [Redis itself](https://github.com/redis/redis/blob/ff3a3577f29a07e965b3b38d07894410d28da499/src/sds.h#L88) so we try to minimise divergence.

I totally understand that. If I manage to get a PR through in Redis itself, will that eventually end up in hiredis? If so, I may start there instead.
 
> We recently switched to using the official Redis package so perhaps something has changed. I need to investigate why they're being skipped

Great! Cheers! Oups, now the macOS test failed too. :-)

michael-grunder:
> If I manage to get a PR through in Redis itself, will that eventually end up in hiredis

Yes for sure

> Great! Cheers! Oups, now the macOS test failed too

The problem is due to Redis 7.0.0 disabling the `DEBUG` command by default.  The fix is pretty simple so should be up shortly.

PR #1066, opened 2022-05-17
Add support for Redis 7's sharded pubsub commands

slact:


bjosv (commented):
I believe we need to handle global channels (SUBSCRIBE) and sharded channels (SSUBSCRIBE) separately. It seems that the `ac->sub.channels` callback is used for both types, but according to the [Redis docs](https://redis.io/commands/sunsubscribe/) they are separate.

slact:
As I understand, there is no difference between the `MESSAGE`s sent to subscribers from `PUBLISH` or `SPUBLISH` commands. Both generate this response:
``` 
1) MESSAGE
2) <channel_name>
3) <message_data>
```

So if you use hiredis to `SUBSCRIBE` and also `SSUBSCRIBE` to channel "foo", when messages arrive it will be impossible to tell if it came from the sharded or the regular channel.

bjosv:
> As I understand, there is no difference between the `MESSAGE`s sent to subscribers from `PUBLISH` or `SPUBLISH` commands. 

Ah, you are right! That's an interesting twist..

bjosv:
There are ongoing discussions regarding this problem in https://github.com/redis/redis/issues/10748

slact:
Oh boy. Well, in the meantime, the official C library does not support the official API.

PR #1085, opened 2022-08-10
Make it possible to set SSL verify mode

stanhu:
If no SSL certificates are provided, many Redis clients default to
disabling SSL peer verification. Previously it was a bit cumbersome to
configure this because the client would either have to reimplement
`redisCreateSSLContext()` or reach into the internals to set the
OpenSSL verify mode.

We can improve the SSL API by introducing a
`redisCreateSSLContextWithOptions()` call that takes into structured
parameters for SSL initialization. This structure contains a verify
mode that is used to set the OpenSSL verify mode.

Relates to https://github.com/redis/hiredis/issues/646

### DIFF
--- a/ssl.c
+++ b/ssl.c
@@ -219,6 +219,25 @@ redisSSLContext *redisCreateSSLContext(const char *cacert_filename, const char *
         const char *cert_filename, const char *private_key_filename,
         const char *server_name, redisSSLContextError *error)
 {
+    redisSSLOptions options = {
+        .cacert_filename = cacert_filename,
+        .capath = capath,
+        .cert_filename = cert_filename,
+        .private_key_filename = private_key_filename,
+        .server_name = server_name,
+        .verify_mode = REDIS_SSL_VERIFY_PEER,
+    };
+
+    return redisCreateSSLContextWithOptions(&options, error);
+}
+
+redisSSLContext *redisCreateSSLContextWithOptions(redisSSLOptions *options, redisSSLContextError *error) {
+    const char *cacert_filename = options->cacert_filename;
+    const char *capath = options->capath;
+    const char *cert_filename = options->cert_filename;
+    const char *private_key_filename = options->private_key_filename;
+    const char *server_name = options->server_name;
+
 #ifdef _WIN32
     HCERTSTORE win_store = NULL;
     PCCERT_CONTEXT win_ctx = NULL;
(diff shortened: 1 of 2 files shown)


michael-grunder:
I'm not opposed to this in principle, as it does provide quite a bit of convenience.

Presently there's quite a bit of boilerplate involved when you're just testing a TLS connection and you don't want/have peer verification.

That said, if we do add a simple options struct, perhaps we should also incorporate other simple options, such as one that would call `SSL_CTX_set_default_verify_paths`?

@yossigo let me know if you have any thoughts.

yossigo:
@michael-grunder I'm not against that either, but slightly concerned about not ending up wrapping too much of OpenSSL. Ideally, we should be able to draw a line and define what capabilities we provide for convenience and what is out of scope and requires users to handle their OpenSSL context.

stanhu:
Another alternative would be to access the `SSL_CTX` pointer by exporting the `redisSSLContext` structure.

yossigo:
@stanhu There's already `redisInitiateSSL()` for those folks who are willing to use OpenSSL directly, what would be the benefit of also exposing `SSL_CTX` from `redisSSLContext`?

stanhu:
@yossigo `redisCreateSSLContext()` does a fair amount of work to set up the SSL paths. Callers would essentially have to duplicate that code if they want to use `redisInitializeSSL()`.

ssl.c
```diff
@@ -219,6 +219,25 @@ redisSSLContext *redisCreateSSLContext(const char *cacert_filename, const char *
...
         const char *server_name, redisSSLContextError *error)
 {
+    redisSSLOptions options;
+
+    options.cacert_filename = cacert_filename;
+    options.capath = capath;
+    options.cert_filename = cert_filename;
+    options.private_key_filename = private_key_filename;
+    options.server_name = server_name;
+    options.verify_mode = REDIS_SSL_VERIFY_PEER;
```
michael-grunder, line 229:
```suggestion
    redisSSLOptions options = {
        .cacert_filename = cacert_filename,
        .capath = capath,
        .cert_filename = cert_filename,
        .private_key_filename = private_key_filename,
        .server_name = server_name,
        .verify_mode = REDIS_SSL_VERIFY_PEER,
    };
```

michael-grunder:
If you don't mind making the aesthetic change, I will get this merged and we can determine later if we want to extend the options at all.

ssl.c
stanhu: Done, thanks.

PR #1091, opened 2022-08-29
Fix some undefined behavior

michael-grunder:
- redisSSLContextError must always be initialized at defintion,
  otherwise when SSL connect succeeds it may not be assigned to a valid error.
  Thus the memory trash remains in the variable, which may sign a misleading error.


michael-grunder:
Closes #960

PR #1094, opened 2022-08-30
Introduce .close method for redisContextFuncs

pizhenwei:
Currently, hiredis supports TCP/SSL/Unix, all of the connection types
use a single FD(int), close() is enough to close a connection. For the
further step, introduce .close method for redisContextFuncs, this
allows to close a complex connection context, for example RDMA.

Signed-off-by: zhenwei pi <pizhenwei@bytedance.com>

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -48,6 +48,7 @@ extern int redisContextUpdateConnectTimeout(redisContext *c, const struct timeva
 extern int redisContextUpdateCommandTimeout(redisContext *c, const struct timeval *timeout);
 
 static redisContextFuncs redisContextDefaultFuncs = {
+    .close = redisNetClose,
     .free_privctx = NULL,
     .async_read = redisAsyncRead,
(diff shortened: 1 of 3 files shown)


michael-grunder:
Changes look fine to me, and at first glance I think we can include them before v2.0.0.

We'll need to bump the `SONAME` but we need to do that for the next version anyway.

pizhenwei:
> Changes look fine to me, and at first glance I think we can include them before v2.0.0.
> 
> We'll need to bump the `SONAME` but we need to do that for the next version anyway.

Hi @michael-grunder 
I'm a little confused, could you please give me any detailed hint about the next step?

michael-grunder:
Nothing needed from your end.  I'll get the changes merged, the only question is if it'll be in the next release, which I'm going to tag soon

michael-grunder:
Merged, thanks!

PR #1096, opened 2022-09-01
Add possibility to prefer IPv6, IPv4 or unspecified

### LINKED ISSUE #696: Resolve IPv6 address takes too much time
We have nss-mdns (http://0pointer.de/lennart/projects/nss-mdns/#documentation) deployed in our system and publish IPv6 address with DNS via avahi-publish. The problem that we realize redis connect via redis-cli takes much time(several seconds). The root cause is hiredis first looks for IPv4 address of the DNS, than looks for IPv6 address. but due to system doesn't publish IPv4 address, It takes time(timeout) to failed to get IPv4 address. 

```
static int _redisContextConnectTcp(redisContext *c, const char *addr, int port,
                                   const struct timeval *timeout,
                                   const char *source_addr) {
    ......
    snprintf(_port, 6, "%d", port);
    memset(&hints,0,sizeof(hints));
    hints.ai_family = AF_INET;
    hints.ai_socktype = SOCK_STREAM;

    /* Try with IPv6 if no IPv4 address was found. We do it in this order since
     * in a Redis client you can't afford to test if you have IPv6 connectivity
     * as this would add latency to every connect. Otherwise a more sensible
     * route could be: Use IPv6 if both addresses are available and there is IPv6
     * connectivity. */
    if ((rv = getaddrinfo(c->tcp.host,_port,&hints,&servinfo)) != 0) {
         hints.ai_family = AF_INET6;
         if ((rv = getaddrinfo(addr,_port,&hints,&servinfo)) != 0) {
            __redisSetError(c,REDIS_ERR_OTHER,gai_strerror(rv));
            return REDIS_ERR;
        }
    }
    ......
}
```
So can hints.ai_family be set AF_UNSPEC instead of try IPv4 and IPv6 in order?

zuiderkwast:
Adds options to select preferred IP stack.

* REDIS_OPT_PREFER_IPV4 = Try IPv4 first, then IPv6 if no IPv4 was found
* REDIS_OPT_PREFER_IPV6 = Try IPv6 first, then IPv4 if no IPv4 was found
* Both flags set = Use getaddrinfo() with AF_UNSPEC. This allows the user or sysadmin to configure preferred IP stacks in /etc/gai.conf.

* None of these flags set = Try IPv4 first, for historical reasons.

Fixes #696.

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -809,6 +809,12 @@ redisContext *redisConnectWithOptions(const redisOptions *options) {
     if (options->options & REDIS_OPT_NOAUTOFREEREPLIES) {
         c->flags |= REDIS_NO_AUTO_FREE_REPLIES;
     }
+    if (options->options & REDIS_OPT_PREFER_IPV4) {
+        c->flags |= REDIS_PREFER_IPV4;
+    }
+    if (options->options & REDIS_OPT_PREFER_IPV6) {
+        c->flags |= REDIS_PREFER_IPV6;
+    }
 
     /* Set any user supplied RESP3 PUSH handler or use freeReplyObject
      * as a default unless specifically flagged that we don't want one. */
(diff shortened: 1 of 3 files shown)


bjosv (approved):
Nice!

michael-grunder:
Awesome, this is a nice improvement.  I'm happy to update our docs.

PR #1098, opened 2022-09-01
uvadapter: reduce number of uv_poll_start calls

michael-grunder:
Internally uv_poll_start iterates over all
attached event handlers to update event mask. It's
quite expensive operation if there many event handlers attached
to a loop.
As redisLibuvEvents.events is a copy of what libuv should see,
we can rely on it to avoid event mask updates.

Signed-off-by: Anton Tiurin <noxiouz@yandex.ru>


michael-grunder:
Thanks, @noxiouz, and apologies for the excessive delay!

PR #1102, opened 2022-09-02
Fix an ASAN issue in redisvFormatCommand()

bjosv:
Calling the standard function `strlen()` with `null` causes undefined behavior.
This is triggered when the string argument is missing for a `%s` command format.

Note:
- The formatted output command the added tests are not changed due to the correction.
- One could argue the importance of this fix.

### DIFF
--- a/hiredis.c
+++ b/hiredis.c
@@ -364,6 +364,7 @@ int redisvFormatCommand(char **target, const char *format, va_list ap) {
             switch(c[1]) {
             case 's':
                 arg = va_arg(ap,char*);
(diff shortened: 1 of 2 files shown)


bjosv:
Interesting.. the added additional `%b` and `%d` tests were affected in some CI targets. Will investigate.

bjosv:
Well, I seems that it's undefined behavior when not giving correct number of arguments to `va_args()` anyway..
Building with -m32 on linux `va_args()` gives an invalid string pointer and a segfault, so I guess there is no remedy in this case.
Closing this PR.

Issue #1100, opened 2022-09-02
 int redisKeepAlive(redisContext *c, int interval)  set in windows invalid


wb2712:
```
int redisKeepAlive(redisContext *c, int **_interval_**) {
    int _**val**_ = 1;
    redisFD fd = c->fd;

    if (setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, &_**val**_, sizeof(val)) == -1){
        __redisSetError(c,REDIS_ERR_OTHER,strerror(errno));
        return REDIS_ERR;
    }

    val = interval;
}
```

```
int win32_setsockopt(SOCKET sockfd, int level, int optname, const void ***_optval_**, socklen_t optlen) {
    int ret = 0;
    if ((level == SOL_SOCKET) && ((optname == SO_RCVTIMEO) || (optname == SO_SNDTIMEO))) {
        const struct timeval *tv = _**optval**_;
        DWORD timeout = tv->tv_sec * 1000 + tv->tv_usec / 1000;
        ret = setsockopt(sockfd, level, optname, (const char*)&timeout, sizeof(DWORD));
    } else {
        ret = setsockopt(sockfd, level, optname, (const char*)optval, optlen);
    }
    _updateErrno(ret != SOCKET_ERROR);
    return ret != SOCKET_ERROR ? ret : -1;
}
```

michael-grunder:
Hi, thanks for the report.

I'm not totally sure what the issue is exactly though.  Are you saying that the calculated timeval is incorrect in windows?

wb2712:
redisKeepAlive     ->  _interval_
not be used  in   win32_setsockopt

michael-grunder:
OK it looks like Windows is special and uses the platform-specific `SIO_KEEPALIVE_VALS` code, which is supposed to be set using [WSAIoctl](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsaioctl).

I don't really have a Windows machine but will attempt to put something together.

wb2712:
int redisKeepAlive(redisContext *c, int **_interval_**) {
    int _**val**_ = 1;
    redisFD fd = c->fd;

    if (setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, &_**val**_, sizeof(val)) == -1)


setsockopt()   used   val,   not  _interval_.

michael-grunder:
I don't really follow.  

[...]

PR #1103, opened 2022-09-02
Use a windows specific keepalive function.

michael-grunder:
While it is possible to toggle `SO_KEEPALIVE` in windows via
`setsockopt`, you have to use `WSAIoctl` to set the interval.

Since `WSAIoctl` can actually do all of this in one call (toggle the
option, and set the corresponding interval), just use that in Windows
and avoid the call to `setsockopt` alltogether.

See: #1100


michael-grunder:
cc @kristjanvalur @bjosv I haven't used a Windows machine in ten years, so if either of you happens to know Windows feel free to let me know if there's something incorrect in this PR.

I just based it on their online docs.

michael-grunder:
GitHub got confused by this commit.  See #1104

PR #1106, opened 2022-09-04
Fix protocol error

michael-grunder:
Implement a fix for a recoverable write failure

See #961 for more discussion

### DIFF
--- a/net.c
+++ b/net.c
@@ -83,14 +83,6 @@ ssize_t redisNetRead(redisContext *c, char *buf, size_t bufcap) {
 
 ssize_t redisNetWrite(redisContext *c) {
     ssize_t nwritten = send(c->fd, c->obuf, sdslen(c->obuf), 0);
-    if (nwritten < 0) {
-        if ((errno == EWOULDBLOCK && !(c->flags & REDIS_BLOCK)) || (errno == EINTR)) {
-            /* Try again later */
-        } else {
-            __redisSetError(c, REDIS_ERR_IO, NULL);
-            return -1;
-        }
-    }
     return nwritten;
 }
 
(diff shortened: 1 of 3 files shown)


michael-grunder:
cc @yossigo Let me know if anything jumps out at you.  Since we're now fully abstracting read/write functions I figured we could just specify `< 0` an unrecoverable error `0` a recoverable one, and `> 1` as success. in the callbacks, which should work even for some exotic future read/write callback.

net.c
```diff
@@ -82,15 +82,19 @@ ssize_t redisNetRead(redisContext *c, char *buf, size_t bufcap) {
...
 ssize_t redisNetWrite(redisContext *c) {
-    ssize_t nwritten = send(c->fd, c->obuf, sdslen(c->obuf), 0);
+    ssize_t nwritten;
+
+    nwritten = send(c->fd, c->obuf, sdslen(c->obuf), 0);
     if (nwritten < 0) {
         if ((errno == EWOULDBLOCK && !(c->flags & REDIS_BLOCK)) || (errno == EINTR)) {
-            /* Try again later */
+            /* Try again */
+            return 0;
```
bjosv, line 91: This match what we have in in ssl.c `redisSSLWrite()`, which seems good.

net.c
```diff
@@ -82,15 +82,19 @@ ssize_t redisNetRead(redisContext *c, char *buf, size_t bufcap) {
...
         } else {
-            __redisSetError(c, REDIS_ERR_IO, NULL);
+            __redisSetError(c, REDIS_ERR_IO, strerror(errno));
```
bjosv, line 93: Should we change `redisNetRead()` to set REDIS_ERR_IO errors with the `strerror(errno)` as well?

yossigo:
@michael-grunder This makes sense.

Issue #767, opened 2020-02-20
hiredis pkg-config file incorrectly ignores multiarch libdir spec'n


pgnd:
building hiredis
```
	cmake \
	 -DCMAKE_INSTALL_PREFIX=/usr/local \
	 -DCMAKE_INSTALL_LIBDIR=lib64 \
	..
	make
	make install
```

*installs* in correct multiarch-specified lib dir, == lib64
```
	ls -al /usr/local/lib*/*hiredis*
		lrwxrwxrwx 1 root root  18 Feb 20 11:31 /usr/local/lib64/libhiredis.so -> libhiredis.so.0.14*
		-rwxr-xr-x 1 root root 88K Feb 20 11:31 /usr/local/lib64/libhiredis.so.0.14*
```

but the installed pkg-config .pc points to incorrect lib path
```
	pkg-config hiredis --libs
		-L/usr/local/lib -lhiredis
```

checking
```
	cat /usr/local/lib64/pkgconfig/hiredis.pc
		prefix=/usr/local
		exec_prefix=${prefix}
		libdir=${exec_prefix}/lib
		includedir=${prefix}/include
		pkgincludedir=${includedir}/hiredis

		Name: hiredis
		Description: Minimalistic C client library for Redis.
		Version: 0.14.0
		Libs: -L${libdir} -lhiredis
		Cflags: -I${pkgincludedir} -D_FILE_OFFSET_BITS=64
```

sourced from
```
	cat /usr/local/src/hiredis/hiredis.pc.in
		prefix=@CMAKE_INSTALL_PREFIX@
		exec_prefix=${prefix}
!!		libdir=${exec_prefix}/lib
		includedir=${prefix}/include
		pkgincludedir=${includedir}/hiredis

		Name: hiredis
		Description: Minimalistic C client library for Redis.
		Version: @PROJECT_VERSION@
		Libs: -L${libdir} -lhiredis
		Cflags: -I${pkgincludedir} -D_FILE_OFFSET_BITS=64

```

incorrectly ignores set multiarch libdir -- and misdirects from actually-installed lib

which breaks any downstream app (e.g., suricata) using pkg-config for hiredis config & linking

michael-grunder:
Hi @pgnd CMake isn't really my thing but does the change I pushed do what you're looking for?

pgnd:
@michael-grunder 

if you fix your typo
```
	prefix=@CMAKE_INSTALL_PREFIX@
[...]

PR #1107, opened 2022-09-06
Fix pkgconfig result for hiredis_ssl

bjosv:
Respect an overridden libdir when installing using CMake.
CMake now generates the `hiredis_ssl.pc` file with the correct result in `libdir` and `Libs`.

See #767 for same fix in `hiredis.pc`


bjosv:
The freebsd-vm in the CI job did not start correctly. It should not depend on this change so the job can probably just be re-run..

PR #1109, opened 2022-09-08
Add timeout support to libhv adapter.

michael-grunder:
Add timeout support to our libhv adapter.

See: #904

### DIFF
--- a/adapters/libhv.h
+++ b/adapters/libhv.h
@@ -67,22 +67,19 @@ static void redisLibhvSetTimeout(void *privdata, struct timeval tv) {
     events = (redisLibhvEvents*)privdata;
     millis = tv.tv_sec * 1000 + tv.tv_usec / 1000;
 
-    /* Libhv disallows zero'd timers so handle that condition first */
     if (millis == 0) {
+        /* Libhv disallows zero'd timers so treat this as a delete or NO OP */
         if (events->timer) {
             htimer_del(events->timer);
             events->timer = NULL;
         }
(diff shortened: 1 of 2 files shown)


michael-grunder:
cc @ithewei This is my attempt to add timeout support to your adapter.  I'm unfamiliar with the library so please let me know if my usage is correct.

ithewei:
> 

You are right, and you can use `htimer_reset` to update timeout multiple times.

ithewei:
> cc @ithewei This is my attempt to add timeout support to your adapter. I'm unfamiliar with the library so please let me know if my usage is correct.
> 
> Edit: I think this logic isn't ideal if users update the timeouts multiple times (as it will add a bunch of timers). Is there a mechanism in `libhv` to find an already existing timer?

You are right, and you can use htimer_reset to update timeout multiple times.

michael-grunder:
@ithewei If you don't mind taking a look at the new timeout logic that would be great.  In reading the `libhv` source code I saw that there are several "built-in" timers in the `hio_t` structure, but it was unclear to me whether I could safely use any of them for this purpose.

@bjosv Let me know if anything jumps out at you.  It should now be possible/efficient to update timeouts arbitrarily when using `libhv`.

bjosv (approved):
Nice!
Maybe example users would get helped by a comment in `debugCallback()` that we expect a NULL reply since DEBUG SLEEP should timeout..
..but it LGTM as is.

PR #1115, opened 2022-09-14
Use __attribute__ when building with Clang on Windows

### LINKED ISSUE #1042: mmintrin.h: error: cannot initialize a parameter ...
Hello,

I have a simple C++ program which includes `hiredis.h` header and then C++ `<future>` header:

```
#include "hiredis.h"
#include <future>
```

When I compile under windows with LLVM clang-cl.exe I get the following compiler errors:

```
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\future:25:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\ppltasks.h:27:
In file included from C:\Program Files\LLVM\lib\clang\11.0.0\include\intrin.h:20:
In file included from C:\Program Files\LLVM\lib\clang\11.0.0\include\x86intrin.h:15:
In file included from C:\Program Files\LLVM\lib\clang\11.0.0\include\immintrin.h:15:
C:\Program Files\LLVM\lib\clang\11.0.0\include\mmintrin.h(67,40): error: cannot initialize a parameter of type '__attribute__((__vector_size__(2 * sizeof(int)))) int' (vector of 2 'int' values) with an rvalue of type '__v2si' (aka 'int')
    return __builtin_ia32_vec_ext_v2si((__v2si)__m, 0);
                                       ^~~~~~~~~~~
C:\Program Files\LLVM\lib\clang\11.0.0\include\mmintrin.h(129,43): error: cannot initialize a parameter of type '__attribute__((__vector_size__(4 * sizeof(short)))) short' (vector of 4 'short' values) with an rvalue of type '__v4hi' (aka 'short')
    return (__m64)__builtin_ia32_packsswb((__v4hi)__m1, (__v4hi)__m2);
                                          ^~~~~~~~~~~~
 an rvalue of type '__v2si' (aka 'int')
    return (__m64)__builtin_ia32_packssdw((__v2si)__m1, (__v2si)__m2);
                                          ^~~~~~~~~~~~
C:\Program Files\LLVM\lib\clang\11.0.0\include\mmintrin.h(189,43): error: cannot initialize a parameter of type '__attribute__((__vector_size__(4 * sizeof(short)))) short' (vector of 4 'short' values) with an rvalue of type '__v4hi' (aka 'short')
    return (__m64)__builtin_ia32_packuswb((__v4hi)__m1, (__v4hi)__m2);
                                          ^~~~~~~~~~~~
...
```

Interestingly enough if I first include `<future>` and then `hiredis.h` it compiles fine.

bjosv:
Since Clang supports `__attribute__` we can avoid disabling it and use packed `sdshdrX` structs when building with Clang on Windows. This also make sure we don't affect subsequent header files that require `__attribute__`.

Note:
Clang attempts to be a drop-in replacement for MSVC and therefor defines `_MSC_VER` to be able to use existing libraries/headers.

Fixes #1042

### DIFF
--- a/sds.h
+++ b/sds.h
@@ -35,9 +35,11 @@
 
 #define SDS_MAX_PREALLOC (1024*1024)
 #ifdef _MSC_VER
-#define __attribute__(x)
 typedef long long ssize_t;
 #define SSIZE_MAX (LLONG_MAX >> 1)
+#ifndef __clang__
+#define __attribute__(x)
+#endif
 #endif
 
 #include <sys/types.h>


bjosv:
As I understand it we don't want to change to much in sds.c/.h, but this is an option to fix #1042

michael-grunder:
> Clang attempts to be a drop-in replacement for MSVC and therefor defines _MSC_VER

It defines `__GNUC__` as well which I recently learned the hard way. :smile:

> As I understand it we don't want to change to much in sds.c/.h, but this is an option to fix https://github.com/redis/hiredis/issues/1042

We may actually have more leeway here because we did a mass renaming of `sds` symbols to `hi_<symbol>` when Redis last upgraded the version of hiredis it builds.

I'll get this merged.

bjosv:
> It defines `__GNUC__` as well which I recently learned the hard way. smile

Ouch :)
It's a bit funny that one can find stackoverflow post like `How to tell Clang to stop pretending to be other compilers?`

PR #1116, opened 2022-09-15
Remove new cmake builtins

michael-grunder:
Rework the CMakeLists.txt file so we don't need to require a very new version of CMake.


bjosv:
Found stuff in #1117 that touches some part of this

PR #1117, opened 2022-09-15
Install on windows fixes

bjosv:
After running a temporary changed CI job (build.yaml, job `windows`)
that builds and installs but using: `ninja -v all install` and a matrix of `CMAKE_BUILD_TYPE`
the following changes were found.
1. The pdb file is not longer created for `hiredis_static` since [#1054](https://github.com/redis/hiredis/pull/1054), so remove installation attempts of the file on build types `Debug` or `RelWithDebInfo`
2. Copy the solution in  [#1054](https://github.com/redis/hiredis/pull/1054) for `hiredis_ssl_static`


bjosv:
One observation is that the debug information is added to the static library indifferent of the build type, i.e even for Release builds. Previously the .pdb debug file were only installed when building with the CMake build type `Debug` or `RelWithDebInfo`.

@kristjanvalur I guess you know Windows best, is this the common way? or should it be changed to only include debug info in `Debug`/`RelWithDebInfo` builds? I think that CMake only adds debug flags in Linux when building for `Debug`/`RelWithDebInfo` (like the -g flag).

kristjanvalur:
Hi, sorry for late reply.
The presence of debug info in the .lib file has nothing to do with what happens during final linking. During linking of a .dll or .exe a .pdb *may* be generated. The necessary data *can* be found in the .lib file, or ignored. This change only means that an extra sidecar file is not needed to be lugged around with the static lib *if* you want debug info in the final.

PR #1122, opened 2022-09-28
CMake corrections for building on Windows

### LINKED ISSUE #1105: The features used are not compatible with the specified minimum version of cmake 3.0 for WIN32.
ADD_COMPILE_DEFINITIONS requires 3.12
TARGET_FILE_BASE_NAME requeres cmake 3.15

bjosv:
- Only use the `/Z7` flag on a Visual Studio compilator.
  This flag is not available on a GNU compilator running on windows, like in a mingw64 setup ([example](https://github.com/Nordix/hiredis-cluster/blob/3c34c78471b8eea331cd08fee662850ca8576d7d/.github/workflows/ci.yml#L139-L159)).

- Replace newer CMake builtins that does not exists in the minimum required version `v3.0.0`.
  `ADD_COMPILE_DEFINITIONS` is replaced with `ADD_DEFINITIONS` to set defines.
  CMake's Visual Studio generator will replace `-D` with `/D` (but all other generators seems not doing the opposite..) 

  Fixes #1105


bjosv:
@SRyabinin @kristjanvalur My testing seems fine, but please check if reasonable.

kristjanvalur:
Looks reasonable.  Btw, I'm not in any way a CMake expert, I have just sort of tried to do my best to make things work :)

michael-grunder:
Merged, thanks!

Also from where I'm sitting you are both CMake gurus :smile:

PR #1133, opened 2022-11-06
Add support for nan in RESP3 double

filipecosta90:
Given https://github.com/redis/redis-specifications/pull/10, addresses https://github.com/redis/redis-specifications/issues/9 in hiredis.


michael-grunder:
Thanks for the submission.

I'm tagging `v1.1.0-rc1` this evening and think this may need to go into a subsequent release.

Once merged, applications that use hiredis will need to account for `nan` whereas presently, that would be treated as an error by hiredis itself.

PR #1147, opened 2022-12-22
CMakeLists.txt: respect BUILD_SHARED_LIBS

ffontaine:
To allow building hiredis on toolchain without dynamic library support, respect standard cmake `BUILD_SHARED_LIBS`:
https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


michael-grunder:
This certainly looks like a nice improvement, but I'm not particularly familiar with CMake.

cc @bjosv @chayim let me know if you see any issues.

chayim:
I like it. Looks like the right approach, and locally it works well.

chayim (approved):

CMakeLists.txt
```diff
@@ -236,10 +262,10 @@ ENDIF()
 IF(NOT DISABLE_TESTS)
     ENABLE_TESTING()
     ADD_EXECUTABLE(hiredis-test test.c)
-    TARGET_LINK_LIBRARIES(hiredis-test hiredis)
+    TARGET_LINK_LIBRARIES(hiredis-test ${HIREDIS_DEFAULT_LIBRARY})
```
bjosv, line 265: I believe the same change is needed in `examples/CMakeLists.txt`, at least for windows.

bjosv (commented):
Nice.
I think there are some small additional changes needed to make it work on windows.
On line 127, something like:
-if (MSVC)
+IF(MSVC AND BUILD_SHARED_LIBS)

CMakeLists.txt
bjosv: ..and a  `HIREDIS_SSL_DEFAULT_LIBRARY`

ffontaine: Indeed, I'll update the PR

CMakeLists.txt
```diff
@@ -103,13 +118,13 @@ set(CPACK_RPM_PACKAGE_AUTOREQPROV ON)
...
 
-INSTALL(TARGETS hiredis hiredis_static
+INSTALL(TARGETS ${HIREDIS_TARGETS}
     EXPORT hiredis-targets
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
-if (MSVC)
+if (MSVC AND BUILD_SHARED_LIBS)
```
bjosv, line 127: I missed two others;
- a similar change for `hiredis_ssl` on line 228
- a non-checked use of `hiredis_ssl` on line 217 (which in legacy is a bit weird, but maybe needed)

ffontaine: Thanks, I updated the PR

bjosv (approved):
LGTM!

michael-grunder:
Thanks everyone, merged!

PR #1151, opened 2023-01-04
Fix wincrypt symbols conflict

hudayou:
With openssl variants such as boring ssl, the conflict symbols are not undefined such as openssl.
Explicit undefine those so it works for both normal openssl and  boring ssl

### DIFF
--- a/ssl.c
+++ b/ssl.c
@@ -40,6 +40,14 @@
 #ifdef _WIN32
 #include <windows.h>
 #include <wincrypt.h>
+#ifdef OPENSSL_IS_BORINGSSL
+#undef X509_NAME
+#undef X509_EXTENSIONS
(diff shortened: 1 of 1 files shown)


michael-grunder:
I've never used boringSSL or wincrypt (I don't have a windows machine), so can you provide a bit more context?

I'm not sure what downstream consequences undefining the symbols might have if we do it globally.  Maybe we could limit the lines to only execute if there is some kind of `IS_BORINGSSL` macro.

Similar issue:
https://bugs.chromium.org/p/boringssl/issues/detail?id=371

hudayou:
Looks like there is one
https://boringssl.googlesource.com/boringssl/+/HEAD/PORTING.md

So I updated the change with that 

```
#ifdef OPENSSL_IS_BORINGSSL
```

PR #1154, opened 2023-01-06
Fix typo

szepeviktor:
Just that.

`lintian` told me to do it.


michael-grunder:
I think this was patched in #1153

szepeviktor:
@michael-grunder Just a couple of minutes!!

PR #1188, opened 2023-05-25
Helper for setting TCP_USER_TIMEOUT socket option

zuiderkwast:
This adds another option in the same style as the TCP KeepAlive feature, i.e. passing the option directly to setsockopt() without storing it in the context, meaning it does not happen automatically after redisReconnect().

It is already possible to achieve this by taking the fd out of the redisContext as `c->fd`, so it is not strictly necessary to add this.

Documentation is added in README also for the keepalive feature.


michael-grunder:
This is great, thanks!

The test failures appear totally unrelated so I'll figure those out in another commit.

PR #1190, opened 2023-05-30
Fix a typo in b6a052f.

yossigo:


michael-grunder:
Thanks!

I love this because I looked at all those lines and didn't see it.

bjosv:
The CI step that builds with CMake is using `-Werror`, but lacks some warning-options that is included in the Makefile.
The CI step that builds using the Makefile finds this issue due to more warning-flags, but didn't fail due to the lack of `-Werror`.
Maybe we should add `-Werror` to the CI-steps that builds using the Makefile too? 

An alternative is update both CMakeLists and Makefile to be similar regarding to warnings (and if warning is an error?), but this might need special handling in the CMakeLists for different compilers, like MSVC (and/or platforms?).

net.c
```diff
@@ -236,7 +236,7 @@ int redisContextSetTcpUserTimeout(redisContext *c, unsigned int timeout) {
     res = -1;
     (void)timeout;
 #endif
-    if (res == -1); {
+    if (res == -1) {
         __redisSetErrorFromErrno(c,REDIS_ERR_IO,"setsockopt(TCP_USER_TIMEOUT)");
         redisNetClose(c);
         return REDIS_ERR;
```
zuiderkwast, line 242: What! That's a disaster. :) It would always return REDIS_ERR.

I should have used this in a test. It would have caught it.

Sorry for this and well spotted @yossigo.

PR #1192, opened 2023-05-30
Add a test for the TCP_USER_TIMEOUT option.

michael-grunder:
See #1188, #1190

### DIFF
--- a/test.c
+++ b/test.c
@@ -409,10 +409,19 @@ static void test_tcp_options(struct config cfg) {
     redisContext *c;
 
     c = do_connect(cfg);
+
     test("We can enable TCP_KEEPALIVE: ");
     test_cond(redisEnableKeepAlive(c) == REDIS_OK);
 
-    disconnect(c, 0);
+#ifdef TCP_USER_TIMEOUT
+    test("We can set TCP_USER_TIMEOUT: ");
+    test_cond(redisSetTcpUserTimeout(c, 100) == REDIS_OK);
+#else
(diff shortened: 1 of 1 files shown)


test.c
```diff
@@ -409,10 +409,19 @@ static void test_tcp_options(struct config cfg) {
...
     test_cond(redisEnableKeepAlive(c) == REDIS_OK);
 
-    disconnect(c, 0);
+#ifdef TCP_USER_TIMEOUT
+    test("We can set TCP_USER_TIMEOUT: ");
+    test_cond(redisSetTcpUserTimeout(c, 100) == REDIS_OK);
+#else
+    test("Setting TCP_USER_TIMEOUT errors when unsupported: ");
+    test_cond(redisSetTcpUserTimeout(c, 100) == REDIS_ERR && c->err == REDIS_ERR_IO);
+#endif
```
zuiderkwast, line 422: I'm thinking that since the implementation is using `__redisSetErrorFromErrno` we should also set `errno = ENOTSUP` before calling `__redisSetErrorFromErrno` in the case where TCP_USER_TIMEOUT is not defined. Otherwise errno has some old value from some previous error.

zuiderkwast (approved):
Thanks for adding the test I should have added.

test.c
michael-grunder: Done.  Good call.

zuiderkwast (approved):

michael-grunder:
cc @yossigo I need a review approval to merge :). Or feel free to merge if you prefer.

yossigo (approved):

PR #1193, opened 2023-06-01
Add -Werror as a default.

yossigo:
Addresses the valid concerns raised in https://github.com/redis/hiredis/pull/1190#issuecomment-1570270521.

The other changes resolve minor issues which fail the current CI on different platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ export REDIS_TEST_CONFIG
 CC:=$(shell sh -c 'type $${CC%% *} >/dev/null 2>/dev/null && echo $(CC) || echo gcc')
 CXX:=$(shell sh -c 'type $${CXX%% *} >/dev/null 2>/dev/null && echo $(CXX) || echo g++')
 OPTIMIZATION?=-O3
-WARNINGS=-Wall -W -Wstrict-prototypes -Wwrite-strings -Wno-missing-field-initializers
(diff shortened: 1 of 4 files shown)


Romain-Geissler-1A:
Normally packages which ensure -Werror also do provide a way to explicitly disable this, as warnings are something "volatile" which can happen with new compilers (including non released ones), and also with LTO combined with warnings that generate quite some false positives, it makes the build quite error prone for packagers.

Would you be ok to add an opt-out option to optionally (and explicitly) turn off warnings for packagers ?

yossigo:
@Romain-Geissler-1A This is a valid point, but can't you already do that by overriding the `WARNINGS` Makefile variable to whatever suits your platform? This is even more powerful than dropping `-Werror,` as you can adjust specific flags as needed.

Romain-Geissler-1A:
This hardly scales. When you package many packages, and each of them decide to go in its own way, it makes it more difficult for packaging maintainers to keep up with upstream changes. Even if it looks trivial for one project (like hiredis), it's for me not a solution.

You only want -Werror in your own CI by the way, or in "development mode" when you built it locally. You don't want -Werror in general when people only consume hiredis and don't intend to develop/fix anything in hiredis. As said, other projects just ensure this -Werror for "development" mode and don't enforce it by default in their builds (or if they do, they provide a way to explicitly disable -Werror). Typically this is what glibc does, with a configure flag `--disable-werror`.

PR #1194, opened 2023-06-05
Ensure functionality without `_MSC_VER` definition

windyakin:
When building with a compiler that is not the Microsoft C/C++ compiler (MSVC), I get the error `error: "_MSC_VER" is not defined`.

I checked the code and it looks like you just want to make sure _MSV_VER is defined, so I changed it from `#if` to `#ifdef`.

### DIFF
--- a/test.c
+++ b/test.c
@@ -78,7 +78,7 @@ static int tests = 0, fails = 0, skips = 0;
 
 static void millisleep(int ms)
 {
-#if _MSC_VER
+#ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


windyakin:
I think the cause of the error is https://github.com/redis/hiredis/pull/1193 (`-Werror`)

windyakin:
@yossigo @michael-grunder please check it?

michael-grunder (approved):
This looks good, and I don't see any way that it's harmful.

PR #1200, opened 2023-06-21
fixed `ISO C90 forbids mixed declarations and code`

homirun:
I got an error `error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]`, so I fixed the ISO C90 violation in net.c.

Specifically, we are attempting to resolve this by moving the definition of `error` to the beginning of the file.

### DIFF
--- a/net.c
+++ b/net.c
@@ -306,11 +306,11 @@ static int redisContextWaitReady(redisContext *c, long msec) {
 
 int redisCheckConnectDone(redisContext *c, int *completed) {
     int rc = connect(c->fd, (const struct sockaddr *)c->saddr, c->addrlen);
+    int error = errno;
     if (rc == 0) {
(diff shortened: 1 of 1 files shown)


bjosv (commented):
As I understand it hiredis is targeted >= C99.

Current makefiles at least has:

https://github.com/redis/hiredis/blob/052f99ab2da451e792a8cd264d61017487eae4a0/Makefile#L270

https://github.com/redis/hiredis/blob/052f99ab2da451e792a8cd264d61017487eae4a0/CMakeLists.txt#L28

homirun:
Sorry!
I was using hiredis from mruby in my environment, but it seems that `-Wdeclaration-after-statement` was enabled in older versions of mruby.

Thanks for taking a look!

Issue #990, opened 2021-10-07
1.0.1 release bumps SONAME unnecessarily


jfeep:
The 1.0.1 release fixes CVE-2021-32765 by changing the internals of a function.  The ABI remains untouched, hence the 1.0.1 release didn't need to change the SONAME.

https://github.com/redis/hiredis/commit/8d1bfac4640fe90cd6f800d09b7f53e886569b98#r57634172

Let's put HIREDIS_SONAME back to 1.0.0 and release as 1.0.2.  This will allow binaries that have linked against /usr/lib64/libhiredis.so.1.0.0 to get the CVE fix with a drop-in upgrade that avoids the need to rebuild programs that link against hiredis.

jfeep:
Agree, for `master`, `SONAME` can move ahead (e.g. to `1.0.3-dev`) allowing room for ABI and API changes as development continues.  Then when the next release time comes, you can determine if ABI or API compatibility is possible and determine what the next release number should be (and whether `SONAME` bump is needed).

#991 should have been targeted at the release/v1.0.2 branch rather than `master`.  Since d4e6f10 fixes the issue on the `release/v1.0.2` branch, I'll close PR #991.

This issue can closed once v1.0.2 gets released.  I'll then look at packaging it for [OpenSUSE Tumbleweed](https://software.opensuse.org/package/hiredis)

jfeep:
Thank you @michael-grunder for fixing the SONAME issue.  v1.0.2 package has been prepared for openSUSE Tumbleweed.

rgacogne:
Thanks @michael-grunder for fixing the issue and for notifying me, much appreciated!

PR #1202, opened 2023-07-02
Version 1.2.0

chayim:


CHANGELOG.md
```diff
@@ -1,20 +1,71 @@
+## [1.2.0](https://github.com/redis/hiredis/tree/v1.1.0) - (2022-11-15)
```
michael-grunder, line 1:
```suggestion
## [1.2.0](https://github.com/redis/hiredis/tree/v1.2.0) - (2023-06-04)
```

hiredis.h
```diff
@@ -48,7 +48,7 @@ typedef long long ssize_t;
 #define HIREDIS_MAJOR 1
 #define HIREDIS_MINOR 1
 #define HIREDIS_PATCH 1
```
michael-grunder, line 50:
```suggestion
#define HIREDIS_MINOR 2
#define HIREDIS_PATCH 0
```

michael-grunder (approved):
This is good, just minor define/date touchups.

ABI compatibility is 100% so there are no issues there.

[hiredis-compat-report.html.zip](https://github.com/redis/hiredis/files/11943176/hiredis-compat-report.html.zip)

michael-grunder:
cc @yossigo Do you mind weighing in on whether or not we should bump the `SONAME` here?  `v1.2.0` has a new async adapter, and one new `redisConnectWithOptions` option, but is 100% compatible (API and ABI) with v1.1.0.

Let me know if I'm missing something, but I think we can maintain the SONAME at v1.1.0 so downstream packages that currently link with it don't need to be rebuilt.

yossigo (dismissed):
@michael-grunder The right thing to do is bump the minor version to indicate we've extended the API/ABI in a backward-compatible manner.

In the case of Linux, this shouldn't require recompilation because shared object dependencies are expressed using the major version alone. IIRC, some other platforms do link against a specific minor version but have a SemVer-aware dynamic linker that will pick newer minor versions.

michael-grunder:
Thanks @yossigo.  The reason I ask is because I ran into this before (see #990).  I'm just worried that we'll immediately get cranky comments from package maintainers.  :smile:

yossigo:
@michael-grunder The complaint was about changing SONAME for no reason since this was an internal fix with no ABI impact. Now, we _are_ extending the ABI, so bumping the minor versions seems justified and required.

BTW, while technically it was not necessary to change the version in 1.0.1, my understanding is that it should not have a negative impact on any system that assumes versioning adheres to SemVer. But I guess this is implementation specific and, on some platforms, _may_ come with a price.

Maybe @jfeep or @rgacogne can chime in and share their perspective with us.

rgacogne:
I'm not an expert on the subject at all, but my understanding is that the `SONAME` is about backward-compatibility, and nothing else, so in my humble opinion it should only be bumped when an existing symbol is removed, or the signature of an existing symbol changed. The Debian documentation [1] seems to agree:

> Every time the shared library ABI changes in a way that could break binaries linked against older versions of the shared library, the SONAME of the library and the corresponding name for the binary package containing the runtime shared library should change. Normally, this means the SONAME should change any time an interface is removed from the shared library or the signature of an interface (the number of parameters or the types of parameters that it takes, for example) is changed. This practice is vital to allowing clean upgrades from older versions of the package and clean transitions between the old ABI and new ABI without having to upgrade every affected package simultaneously.

> The SONAME and binary package name need not, and indeed normally should not, change if new interfaces are added but none are removed or changed, since this will not break binaries linked against the old shared library.

Now in theory changing the minor version could be done without triggering a rebuild of binaries linked against `libhiredis.so`, but only if the binaries have been linked against the `libhiredis.so.MAJOR` version (this can be checked by looking at the `NEEDED` section in the binary via, for example, `objdump -p <binary> | grep -F NEEDED | gerp -F libhiredis`). Right now it looks like `hiredis`'s `Makefile` does not generate the symbolic link for the `MAJOR` `SONAME`, but only for the `MINOR`:
```Makefile
        $(INSTALL) $(DYLIBNAME) $(INSTALL_LIBRARY_PATH)/$(DYLIB_MINOR_NAME)
        cd $(INSTALL_LIBRARY_PATH) && ln -sf $(DYLIB_MINOR_NAME) $(DYLIBNAME)
```

I have not looked at what other distributions are doing, but on Arch it means we currently have `/usr/lib/libhiredis.so` and `/usr/lib/libhiredis.so.1.1.0`, and packages are linked against `/usr/lib/libhiredis.so.1.1.0` and will require a rebuild for the slightest `SONAME` change. For example for `Unbound`:

```shell
$ objdump -p /usr/bin/unbound | grep -F NEEDED | grep -F libhiredis
  NEEDED               libhiredis.so.1.1.0 
```

Therefore I would personally prefer the `SONAME` to remain untouched, since the ABI remains fully backward-compatible, and I would suggest adding a symbolic link for `libhiredis.so.MAJOR` in the next release so you have more options in later releases. But I will of course do the rebuild of affected Arch packages if needed :)

[1]: https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#run-time-shared-libraries

chayim:
@rgacogne Happy to revert the SONAME side. To be fair, I thought that was part and parcel of these minor bumps. Reverted SONAME here back to 1.1.1-dev to match.

hiredis.h
```diff
@@ -46,8 +46,8 @@ typedef long long ssize_t;
 #include "alloc.h" /* for allocation wrappers */
 
 #define HIREDIS_MAJOR 1
-#define HIREDIS_MINOR 1
-#define HIREDIS_PATCH 1
+#define HIREDIS_MINOR 2
+#define HIREDIS_PATCH 0
 #define HIREDIS_SONAME 1.1.1-dev
```
michael-grunder, line 51: We want to maintain the same `SONAME` as exists in v1.1.0

```suggestion
#define HIREDIS_SONAME 1.1.0
```

michael-grunder:
Thanks @rgacogne that's very useful info.

I vote we maintain the `SONAME` at `1.1.0`.  Unless there are any objections we can do that and then tag.

hiredis.h
michael-grunder: I realized this may not be perfectly clear.  We just need to make sure the `SONAME` is unchanged from `1.1.0` in the actual tag commit.

yossigo:
Thanks @rgacogne, I actually did not realize hiredis does not symlink the major version - this makes a big difference. I agree we should avoid bumping the minor version in this case, and created #1204 to help with that in the future.

michael-grunder:
@chayim We should make sure #1204 gets into v1.2.0

chayim:
@michael-grunder I see it just got merged... Assuming we're good with this, and there are no changes as discussed... 

1. approve + merge
2. release-drafter -> tag

WDYT?

michael-grunder (approved):

chayim:
@michael-grunder last bit of this, to ensure no scope creep. Confirming merge of #1205 in here (please thumbs up/down) and then that gets merged, and this plan followed.

Thumbs down if you disagree with the merge proposal, and then we release as is. Last step @yossigo before release.

PR #1204, opened 2023-07-06
Install major version symlink of shared objects.

yossigo:
This change addresses the issue discussed in #1202 and should make it possible in the future to update minor versions without requiring re-linking binaries.


michael-grunder:
Thanks, this will make things nicer going forward.

 I'll merge after we tag v1.2.0

michael-grunder (approved):

yossigo:
I don't see a problem getting it into 1.2.0.

PR #1205, opened 2023-07-09
Remove support in deprecated TLS versions 1.0 and 1.1

valentinogeron:


valentinogeron:
@michael-grunder @chayim @yossigo - I saw you plan to release a new hiredis version (1.2.0) soon, WDYT on adding it into it?

ssl.c
```diff
@@ -256,13 +256,25 @@ redisSSLContext *redisCreateSSLContextWithOptions(redisSSLOptions *options, redi
     if (ctx == NULL)
         goto error;
 
-    ctx->ssl_ctx = SSL_CTX_new(SSLv23_client_method());
+    const SSL_METHOD *ssl_method;
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
```
chayim, line 260: Perhaps move the check var (0x10100000L) into a const - to keep it in one place?

michael-grunder:
Can we include it with only a minor version bump?

chayim:
@michael-grunder Fully supportive of 1.2.0ing this.

chayim (approved):

yossigo:
> Can we include it with only a minor version bump?

I don't think there's an obvious answer since we do not break the API or any documented guarantee about TLS compatibility.

Practically speaking, it's unlikely that this will have a negative impact since TLSv1.2 has been around for a long time. So, I lean towards saying the security value here outweighs the compatibility issue, being vague as it is.

PR #1211, opened 2023-07-13
Stop hardcoding -std=c99 explicitly.

Romain-Geissler-1A:
Gcc since gcc >= 5 uses C11 (which superseeds C99). Clang uses C11 (or C17) since at least Clang >= 3.3 Gcc 5 was released in april 2015.

The reasoning here is that distros (packagers) already force some default dialect. It's not super important in C, but in C++ it matters more to have all the libraries ideally built with the very same dialect. Now compilers have adopted the idea that when a standard is "well implemented" and was "battle-tested" a bit, it shall be defined as the default dialect (typically couple of years after it was initially introduced). Usually libraries vendor are advised to check that the default feature (here C99) are used by default, instead of forcing them. 

Also if really we wanted to enforce this C99 flag, it should have been done in the global flag variable rather than here in the middle of the makefile.

In our case, our build system ensures C17 for now (by defininng it in the environment $CFLAGS)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -267,7 +267,7 @@ check: hiredis-test
 	TEST_SSL=$(USE_SSL) ./test.sh
 
 .c.o:
-	$(CC) -std=c99 -c $(REAL_CFLAGS) $<
+	$(CC) -c $(REAL_CFLAGS) $<
 
 clean:
 	rm -rf $(DYLIBNAME) $(STLIBNAME) $(SSL_DYLIBNAME) $(SSL_STLIBNAME) $(TESTS) $(PKGCONFNAME) examples/hiredis-example* *.o *.gcda *.gcno *.gcov


michael-grunder:
The issue I see with removing `-std=c99` is that it will lower the quality of diagnostics we get when doing non-standard things.

I think Redis is now targeting `-std=c11` so perhaps that is an option?

cc @yossigo What do you think.

Romain-Geissler-1A:
You mean you are more concerned by the fact that by default gcc/clang use the `-std=gnuN` dialect rather than `-std=cN` dialect ? In that case, since you already use `-Werror` by default, may I suggest you add `-pendantic` instead, to catch your usage of the GNU extension ? And keep the default dialect of the compiler untouched.

michael-grunder:
We do use pedantic, but that works in conjunction with the `-std=c99` flag to enforce ISO C99 compliance in the hiredis codebase.  You're completely correct that this is now an ancient standard, but hiredis is in use on a wide variety of systems including esoteric ones like AIX.

We've actually had people complain (more than once) that we don't target c89.

Romain-Geissler-1A:
Ok so I guess I can just close this pull request then ? Even your initial suggestion of maybe using `c11` doesn't seem to be acceptable if people want to target old platforms.

michael-grunder:
Yeah, I don't think we can remove specifying the standard.  We may be able to get away with bumping it, but I'm not even sure about that.

PR #1213, opened 2023-07-19
Retry poll(2) if we are intterupted.

### LINKED ISSUE #1206: EINTR on connect.
I ran into an issue where the [poll](https://github.com/redis/hiredis/blob/d543baba671bc4609d27e2905de0d8d987f708fe/net.c#L283) syscall is interrupted, which causes hiredis to fail with this error:

```poll(2): Interrupted system call```

In other parts of hiredis, we detect `EINTR` and retry the syscall and was wondering if we should do something similar here.

To maintain backward compatibility we could make it one of our `redisOptions` options (e.g. "ignore EINTR on connect").

cc @yossigo @chayim  @bjosv What do you guys think?

Absolutely no rush on this, from the client side it's simple enough to just retry the connection, but it may be worth adding to hiredis.

michael-grunder:
This commit adds logic to retry our poll call when waiting for the connection to complete, in the event that we are interrupted by a signal.

Additionally we do some simple bookkeeping to keep track of the overall timeout specified by the user.

Fixes #1206

### DIFF
--- a/net.c
+++ b/net.c
@@ -41,6 +41,7 @@
 #include <stdio.h>
 #include <limits.h>
 #include <stdlib.h>
+#include <time.h>
 
 #include "net.h"
 #include "sds.h"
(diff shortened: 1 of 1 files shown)


michael-grunder:
cc @yossigo 

Part of me thinks this is overkill but OTOH it seems like a good idea to respect the users's total requested connection timeout in the context of an interrupted sycall.

yossigo (approved):
@michael-grunder LGTM. I agree this is somewhat overkill but I think it's the best way to handle it given the constraints of current API and our desire to minimize breaking changes. I think we should clearly document this behavior change, by the way.

PR #1223, opened 2023-09-01
Fixed crash when

kaegoorn:
Fixed crash when __redisAsyncFree is called if redisLibuvAttach has not completed

`       context = redisAsyncConnect(host, port);

        if (redisLibuvAttach(context, loop) != REDIS_OK)
             redisAsyncFree(context); // <--- crash
`

### DIFF
--- a/adapters/libuv.h
+++ b/adapters/libuv.h
@@ -130,6 +130,9 @@ static void redisLibuvSetTimeout(void *privdata, struct timeval tv) {
 static void redisLibuvCleanup(void *privdata) {
     redisLibuvEvents* p = (redisLibuvEvents*)privdata;
 
+    if (p == NULL)
+        return;
+    
     p->context = NULL; // indicate that context might no longer exist
     if (p->timer.data) {
         uv_close((uv_handle_t*)&p->timer, on_timer_close);


adapters/libuv.h
```diff
@@ -130,6 +130,9 @@ static void redisLibuvSetTimeout(void *privdata, struct timeval tv) {
 static void redisLibuvCleanup(void *privdata) {
     redisLibuvEvents* p = (redisLibuvEvents*)privdata;
 
+    if (p == NULL)
```
AgranatMarkit, line 133: 1. It's not a good practice, to add such checks, because they could affect performance.
2. You couldn't check redundant cleanups in client code, if it's needed.
3. It's better to make early return before casting on line 131.

kaegoorn: Hi Mark
1. This is a best practice to check variables
2. This is internal function calls from driver
3. May be

P.s. Марк, ты вообще офигел :)
Я сейчас вспомню Серафим и всякое разное :)

AgranatMarkit: As i see, it's better to add check for nullability [here](https://github.com/redis/hiredis/blob/869f3d0ef1513dd0258ad7190c9914df16dcc4a4/async_private.h#L52), because there are [exist](https://github.com/search?q=repo%3Aredis%2Fhiredis%20ev.cleanup&type=code) another adapters, and their cleanup functions will be also called with NULL `privdata`.

AgranatMarkit (approved):

AgranatMarkit (approved):

Issue #1231, opened 2023-10-06
CMake sets incorrect version for library


kendamaskareivis:
When compiling this library, incorrect version is linked to shared object file. I pulled latest 1.2.0 tag, it's creating 1.1.0 instead of 1.2.0.
https://github.com/redis/hiredis/blob/master/CMakeLists.txt#L53
https://github.com/redis/hiredis/blob/master/CMakeLists.txt#L167

Plus it would be very nice to also add in symlinking to major release so when compiling this library we would get .so .so.1 .so.1.x.x instead of currently .so .so.1.x.x

```
# Proper symlinks with correct Cmakefile
lrwxrwxrwx 1 root root     15 okt  6 13:33 libhiredis.so -> libhiredis.so.1
lrwxrwxrwx 1 root root     19 okt  6 13:33 libhiredis.so.1 -> libhiredis.so.1.2.0
-rwxrwxr-x 1 root root 106296 okt  6 13:33 libhiredis.so.1.2.0

# What current project Cmakefile is doing
lrwxrwxrwx 1 root root     19 okt  6 13:44 libhiredis.so -> libhiredis.so.1.1.0
-rwxrwxr-x 1 root root 106296 okt  6 13:44 libhiredis.so.1.1.0
```

This is caused because of improper CMake configuration and wrong `hiredis.h` header `.so (shared object)` definitions

truongan07:
I get error ./hiredis-test: error while loading shared libraries: libhiredis.so.1.1.0: cannot open shared object file: No such file or directory on hiredis-1.2.0. How to fix ?

uglide:
Fixed in 1.3.0

PR #1232, opened 2023-10-06
 #1231 Resolve

kendamaskareivis:
See  #1231

# Changes
- change in `hiredis.h` headers
- changes in `Cmakelists.txt` file


michael-grunder:
Thanks for the PR.  This does improve things in terms of CMake.

The only issue I see is that the change removes any distinction between v1.2.0 proper and the prerelease development branch.

Also the reason we're creating v1.1.0 for v1.2.0 is that v1.2.0 is still API/ABI compatible with v1.1.0 and we were avoiding breaking changes for existing users of v1.1.0

kendamaskareivis:
Forgive me if I'm in the wrong here @michael-grunder , but it seems silly to me that feature updates are actually breaking changes, something to me doesn't add up here. Well anyway, I'd hope you'll reconsider this PR.

michael-grunder:
Appologies if I wasn't clear, but I will try to get this merged :smile:

I just need to make sure we don't break existing code is all.

The hiredis SONAME has been a bit of a sticking point for legacy reasons.  You can [read some of the discussion around this here](https://github.com/redis/hiredis/pull/1202).

uglide:
Superseeded by https://github.com/redis/hiredis/pull/1296

PR #889, opened 2020-10-18
Add support for Wincert CA store in Windows

michael-grunder:

PR #1235, opened 2023-11-07
pull certificates from CA store on Windows, additionally to current Root system certificate store added with #889

masariello:
#889 added support for pulling SSL certificates from the Root Windows certificate store. Turns out one actually needs both the Root and CA stores, so this adds the latter

### DIFF
--- a/ssl.c
+++ b/ssl.c
@@ -288,27 +288,32 @@ redisSSLContext *redisCreateSSLContextWithOptions(redisSSLOptions *options, redi
     if (capath || cacert_filename) {
 #ifdef _WIN32
         if (0 == strcmp(cacert_filename, "wincert")) {
-            win_store = CertOpenSystemStore(NULL, "Root");
-            if (!win_store) {
-                if (error) *error = REDIS_SSL_CTX_OS_CERTSTORE_OPEN_FAILED;
-                goto error;
-            }
-            X509_STORE* store = SSL_CTX_get_cert_store(ctx->ssl_ctx);
-            while (win_ctx = CertEnumCertificatesInStore(win_store, win_ctx)) {
-                X509* x509 = NULL;
-                x509 = d2i_X509(NULL, (const unsigned char**)&win_ctx->pbCertEncoded, win_ctx->cbCertEncoded);
-                if (x509) {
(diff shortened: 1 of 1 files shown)


michael-grunder:
Thanks for the PR.  I can go through the code in the next couple of days but I don't have a Windows machine to verify that the logic works.

masariello:
No worries. It's quite tricky to test. But I did it at work and it did
solve the SSL cert verification I was having. And there are no other
relevant cert stores on WIndows. So this should be it.

Thank you

Il giorno lun 13 nov 2023 alle ore 23:53 Michael Grunder <
***@***.***> ha scritto:

masariello:
@michael-grunder any chance we can merge this? The patch has been running in production since this MR was created :)

michael-grunder:
I think the Redis folks have been taking a more active role in hiredis so I'd see what @uglide thinks.

I pulled the logic into my fork and [tweaked it a bit](https://github.com/michael-grunder/hiredis/commits/ca-wincert/). Mostly just to protect against a NULL deref of `cacert_filename` and a possible double free of the cert itself.

I think there is one blocker though in that this changes the default "wincert" behavior to load the Root and CA store. I am not an expert in OpenSSL by any means but this seems like something users need to opt-in to.

masariello:
> I think the Redis folks have been taking a more active role in hiredis so I'd see what @uglide thinks.
> 
> I pulled the logic into my fork and [tweaked it a bit](https://github.com/michael-grunder/hiredis/commits/ca-wincert/). Mostly just to protect against a NULL deref of `cacert_filename` and a possible double free of the cert itself.
> 
> I think there is one blocker though in that this changes the default "wincert" behavior to load the Root and CA store. I am not an expert in OpenSSL by any means but this seems like something users need to opt-in to.

It would be quite ok to introduce a new "wincert_with_ca" for client code to opt in to load the CA store

All the other changes in your branch look great and cleaner. I d be quite happy to release your changes with the additional "wincert_with_ca"

masariello:
@michael-grunder any chance we can get this PR merged? is there anything I can help with? do you want me to merge your `ca-wincert` branch into this PR?

PR #1257, opened 2024-04-08
Makefile: correctly handle version suffixes on macOS

ZhongRuoyu:
On macOS, the version suffix of a shared library comes before the `.dylib` suffix, which is different from the convention on other systems. Currently, the Makefile only overrides `DYLIB_MINOR_NAME` for macOS, but does not handle the other similar variables, namely `DYLIB_MAJOR_NAME`, `SSL_DYLIB_MINOR_NAME`, and ``SSL_DYLIB_MAJOR_NAME`.

This PR fixes the issue by overriding all these variables for macOS.


ZhongRuoyu:
Gentle ping @michael-grunder -- mind having a look? It'd be nice to have this shipped, and we can apply the fix to Homebrew's hiredis.

michael-grunder (approved):

michael-grunder:
As far as I can tell this change is correct. Thanks!

ZhongRuoyu:
Great, thank you @michael-grunder!

PR #1274, opened 2024-08-06
CentOS 7 is EOL

michael-grunder:
CentOS 7 reached EOL of even maintinance updates on 2024-06-30.


michael-grunder:
The macOS failure isn't related to this PR. I will need to figure out what's going on there.

Looks openssl related

PR #1283, opened 2025-02-04
Do not explicitly initialize openSSL in newer versions

crrodriguez:
Recentish openSSL versions have implicit initialization and is better to let it be, unless you have something special to do.

### DIFF
--- a/ssl.c
+++ b/ssl.c
@@ -167,8 +167,8 @@ static int initOpensslLocks(void) {
 
 int redisInitOpenSSL(void)
 {
-    SSL_library_init();
 #ifdef HIREDIS_USE_CRYPTO_LOCKS
+    SSL_library_init();
     initOpensslLocks();
 #endif
 


michael-grunder:
Only read a bit about the deprecation in newer versions, but I'm not sure why we would want this controled by `HIREDIS_USE_CRYPTO_LOCKS`.

Wouldn't it be more idiomatic to do something like:

```c
// Assuming this is the correct version
#if OPENSSL_VERSION_NUMBER < 0x10100000L
    SSL_library_init();
#endif
```

crrodriguez:
I know it looks odd.. but if you need the crypto locks you need initialization..else you don't.  It was the shortest possible change. 
It can be made version specific if you think it is better.

michael-grunder:
Oh I see now, you're totally correct. 

I assumed `HIREDIS_USE_CRYPTO_LOCKS` was user toggled, but it's simply this:

```c
#if OPENSSL_VERSION_NUMBER < OPENSSL_1_1_0
#define HIREDIS_USE_CRYPTO_LOCKS
#endif
```

So this change is correct.

michael-grunder (approved):

Issue #1207, opened 2023-07-11
Proposal for CHANGES change


chayim:
Given how we now handle CHANGES< and using the release-drafter, we have a chicken and egg issue. The release drafter handles our release notes now. But the merge for a [VERSION](https://github.com/redis/hiredis/pull/1202) is responsible for updating the CHANGES file.   We can continue to _manually_ update the CHANGES file prior to release, but that's less than ideal. We can ask/continue to ask the community to update the CHANGES file, but that's a different barrier to entry.

Proposal: 
- Merges to master update the CHANGES file post-merge. This means a CI run on master, after something has been pushed to master.
- Changes matching some sort of VERSIONING title updates CHANGES, specifically to insert the version number and date, maintaining what we have. 
- Optional (I don't personally see the value): We can pre-include a link to the release notes that will be generated for the version.
- Release are just tags, via the drafter. No changes.

Downside
- Some people dislike automatic updates to master (I'm not one of them)
- The formatting present from the release-drafter wont' be present.

@yossigo @michael-grunder WDYT?

chayim:
@uglide WDYT?

chayim:
The only kicker is that means master `git log` is polluted with 'UPDATING CHANGES' type merges. The flip side is that we tried doing this in branches elsewhere, as parts of PRs - and frankly it makes merging awful (almost always a conflict, due to multiple merges).  Still - I'm pro

chayim:
@dvora-h so that you see this

uglide:
Resolved by https://github.com/redis/hiredis/pull/1294

PR #1294, opened 2025-04-09
Use GitHub Releases to track changelogs

uglide:
The current approach is a huge waste of time since we are just copy-pasting from GitHub releases and [have been discussed in length before. ](https://github.com/redis/hiredis/issues/1207)

As a first step, to simplify release process I suggest dropping this approach altogether and point to the GitHub releases.


StefanPalashev (approved):
Thanks, @uglide! This would simplify the release process.

PR #1296, opened 2025-04-10
Change the version of the library to 1.3

StefanPalashev:
Change the `CMakeLists.txt` to produce a _symlink_ with a generic name that points to the latest major version.

The build output will now look like this:
```
libhiredis.so -> libhiredis.so.1*
libhiredis.so.1* -> libhiredis.so.1.3.0*
libhiredis.so.1.3.0*

```
This would allow for seamless minor version updates.


CMakeLists.txt
```diff
@@ -51,9 +51,13 @@ ADD_LIBRARY(hiredis::hiredis ALIAS hiredis)
 set(hiredis_export_name hiredis CACHE STRING "Name of the exported target")
 set_target_properties(hiredis PROPERTIES EXPORT_NAME ${hiredis_export_name})
 
+# Extract the first part of the HIREDIS_SONAME (major version) and use it for SOVERSION
+STRING(REGEX REPLACE "^([0-9]+)\\..*" "\\1" HIREDIS_SONAME_MAJOR "${HIREDIS_SONAME}")
```
michael-grunder, line 55: Maybe we actually drop `HIREDIS_SONAME` from `hiredi.h` and just use `${HIREDIS_MAJOR}` in here where appropriate?

CMakeLists.txt
```diff
@@ -169,7 +173,8 @@ IF(ENABLE_SSL)
     SET_TARGET_PROPERTIES(hiredis_ssl
         PROPERTIES
         WINDOWS_EXPORT_ALL_SYMBOLS TRUE
-        VERSION "${HIREDIS_SONAME}")
+        VERSION "${VERSION}"
+        SOVERSION "${HIREDIS_SONAME_MAJOR}")
```
michael-grunder, line 177:
```suggestion
        SOVERSION "${HIREDIS_MAJOR}")
```

michael-grunder (commented):
I did some searching around and it seems like very few libraries actually define the `SONAME` in their headers like we do in hiredis.

Maybe the way to go is to drop the soname define in hiredis.h and just use the major version to construct the `SONAME` in both `CMakeLists.txt` and `Makefile`?

That said, I pulled this change and built hiredis and it works well, so if we don't want to remove the SONAME this change is fine with me.

Makefile
uglide, line 1: I think we should revert the changes in this file to make it releasable as a patch release.

hiredis.h
```diff
@@ -48,7 +48,6 @@ typedef long long ssize_t;
 #define HIREDIS_MAJOR 1
 #define HIREDIS_MINOR 3
 #define HIREDIS_PATCH 0
-#define HIREDIS_SONAME 1.3.0
```
uglide: The same as above

uglide (changes requested):

michael-grunder (approved):
LGTM

Makefile
michael-grunder: Right good point because of the SONAME BC.

I guess we can only fix it in 2.0

uglide (approved):

StefanPalashev:
All of the comments have been resolved by applying the suggested changes.

PR #1298, opened 2025-04-22
Define ssize_t as intptr_t in Windows

feihongmeilian:
After checking, the modification typedef intptr_t ssize_t; is more suitable as it self-adapts to the platform and won't cause conflicts when combined with asynchronous libuv under Win32.

### DIFF
--- a/hiredis.h
+++ b/hiredis.h
@@ -39,7 +39,7 @@
 #include <sys/time.h> /* for struct timeval */
 #else
 struct timeval; /* forward declaration */
-typedef long long ssize_t;
+typedef intptr_t ssize_t;
 #endif
 #include <stdint.h> /* uintXX_t, etc */
 #include "sds.h" /* for sds */
--- a/sds.h
+++ b/sds.h
@@ -35,8 +35,8 @@
 
 #define SDS_MAX_PREALLOC (1024*1024)
 #ifdef _MSC_VER
-typedef long long ssize_t;
-#define SSIZE_MAX (LLONG_MAX >> 1)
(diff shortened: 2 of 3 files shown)


michael-grunder (approved):
I don't have or use a Windows machine but from some quick reading it appears this is more correct.

`WIN32` -> `intptr_t` -> `__int32`
`WIN64` -> `intptr_t` -> `__int64`

Whereas `long long` is 64 bits on WIN32.

feihongmeilian:
Yes, according to my research, the C99 standard specifies that long long occupies at least 8 bytes on 32-bit machines.
However, on ​Win32, ssize_t should inherently occupy ​4 bytes, whereas defining it as long long would force it to ​8 bytes, which violates the semantic meaning of ssize_t. Additionally, this discrepancy could cause conflicts when integrating ​libuv​ for asynchronous operations. Below is libuv's definition for reference:[https://github.com/libuv/libuv/blob/v1.x/include/uv/win.h#L26](url)

feihongmeilian:
This problem surfaced when I was compiling redis-plus-plus. I hope my program can run on Win64, Win32, and Linux. However, on Win32, due to the different definitions of ssize_t in hiredis and libuv, redis-plus-plus fails to compile.
Meanwhile, the information I've looked up shows that libuv's definition better conforms to the semantics of ssize_t, so I submitted this PR.

michael-grunder:
Sorry if it wasn't clear, I mean this PR seems correct :smile:

michael-grunder:
cc @uglide Let me know if you have any issues. Seems good to me.

PR #1301, opened 2025-05-22
Apply FD_CLOEXEC on sockets

lennyly:
This PR sets the `FD_CLOEXEC` flag on all sockets created by hiredis to prevent file descriptor leakage across `fork()` abd `exec()` boundaries.

When a program using hiredis forks, the file descriptors for sockets are inherited by the child process unless `FD_CLOEXEC` is set. This can lead to hard to diagnose issues, especially in environments that use epoll or changelist-based polling optimizations.

In one such case, the parent process initiated a reconnection, which reused the same file descriptor. However because the original socket object was still "alive" due to being duplicated in the child process, epoll continued to monitor the old underlying file object associated with the same file descriptor. Since epoll watches file objects, not just file descriptors, this caused the epoll instance to incorrectly associate events with the wrong socket.

Under a changelist optimization, this is particularly problematic because `epoll_ctl()` calls can be skipped, and the internal epoll state may remain stale. As a result, the event loop may enter a tight loop, leading to 100% CPU usage due to busy looping.

By marking sockets with `FD_CLOEXEC`, we ensure that the file descriptor is automatically closed in any `exec()`'d child process, allowing the file object to be fully cleaned up and avoiding the above issue.

### DIFF
--- a/net.c
+++ b/net.c
@@ -127,6 +127,14 @@ static int redisCreateSocket(redisContext *c, int type) {
         return REDIS_ERR;
     }
     c->fd = s;
+
+    /* Prevent file descriptor from leaking to child processes */
+    if (fcntl(s, F_SETFD, FD_CLOEXEC) == -1) {
+        __redisSetErrorFromErrno(c,REDIS_ERR_IO,"fcntl(FD_CLOEXEC)");
+        redisNetClose(c);
+        return REDIS_ERR;
+    }
+
     if (type == AF_INET) {
         if (redisSetReuseAddr(c) == REDIS_ERR) {
             return REDIS_ERR;
@@ -517,6 +525,16 @@ addrretry:
             continue;
 
         c->fd = s;
+
+#ifndef _WIN32
+        /* Prevent file descriptor from leaking to child processes */
+        if (fcntl(s, F_SETFD, FD_CLOEXEC) == -1) {
+            __redisSetErrorFromErrno(c,REDIS_ERR_IO,"fcntl(FD_CLOEXEC)");
+            redisNetClose(c);
+            goto error;
+        }
+#endif
+
         if (redisSetBlocking(c,0) != REDIS_OK)
             goto error;
         if (c->tcp.source_addr) {


net.c
```diff
@@ -127,6 +127,14 @@ static int redisCreateSocket(redisContext *c, int type) {
         return REDIS_ERR;
     }
     c->fd = s;
+
+    /* Prevent file descriptor from leaking to child processes */
+    if (fcntl(s, F_SETFD, FD_CLOEXEC) == -1) {
```
michael-grunder, line 132: Need to wrap this in an `#ifndef _WIN32` type check.

michael-grunder:
This seems like a nice improvement. I guess the only question is whether or not it should be set unconditionally or optionally like `TCP_KEEPALIVE`.

I'm pretty sure `FD_CLOEXEC` is basically everywhere but I'm thinking of exotic systems or very niche use cases. Hiredis is absolutely everywhere so is there an edge case where someone would not want this flag set?

net.c
michael-grunder: Yeah that's even better.

lennyly:
To support all use cases I've added an options flag, and also switched to `SOCK_CLOEXEC` so that the option will be set atomically in `socket()` instead of later on, thus avoiding the risk of having another thread fork in between `socket()` and `fcntl()`.

collinfunk:
> To support all use cases I've added an options flag, and also switched to `SOCK_CLOEXEC` so that the option will be set atomically in `socket()` instead of later on, thus avoiding the risk of having another thread fork in between `socket()` and `fcntl()`.

Yes, that is much better. And they are hidden behind `#ifdef SOCK_CLOEXEC`. Thanks!

One thing. This:
```c
#ifdef SOCK_CLOEXEC
    int flags = SOCK_STREAM;
    if (c->flags & REDIS_OPT_SET_SOCK_CLOEXEC) {
        flags |= SOCK_CLOEXEC;
    }
    if ((s = socket(type, flags, 0)) == REDIS_INVALID_FD) {
        __redisSetErrorFromErrno(c, REDIS_ERR_IO, NULL);
        return REDIS_ERR;
    }
#else
    if ((s = socket(type, SOCK_STREAM, 0)) == REDIS_INVALID_FD) {
        __redisSetErrorFromErrno(c,REDIS_ERR_IO,NULL);
        __redisSetErrorFromErrno(c, REDIS_ERR_IO, NULL);
        return REDIS_ERR;
    }
#endif /* SOCK_CLOEXEC */
```
Can be made shorter by doing:
```c
   int flags = SOCK_STREAM;
#ifdef SOCK_CLOEXEC
    if (c->flags & REDIS_OPT_SET_SOCK_CLOEXEC) {
        flags |= SOCK_CLOEXEC;
    }
#endif /* SOCK_CLOEXEC */
    if ((s = socket(type, flags, 0)) == REDIS_INVALID_FD) {
        __redisSetErrorFromErrno(c, REDIS_ERR_IO, NULL);
        return REDIS_ERR;
    }
```
But will let @michael-grunder what he thinks.

michael-grunder:
Yeah I think that shorter snippit is good. Less duplication.

lennyly:
Good feedback, thanks!

net.c
```diff
@@ -122,11 +122,20 @@ static int redisSetReuseAddr(redisContext *c) {
...
     redisFD s;
-    if ((s = socket(type, SOCK_STREAM, 0)) == REDIS_INVALID_FD) {
-        __redisSetErrorFromErrno(c,REDIS_ERR_IO,NULL);
+    int flags = SOCK_STREAM;
+
+#ifdef SOCK_CLOEXEC
+    if (c->flags & REDIS_OPT_SET_SOCK_CLOEXEC) {
+        flags |= SOCK_CLOEXEC;
+    }
+#endif /* SOCK_CLOEXEC */
```
michael-grunder, line 131:
```suggestion
#endif
```

net.c
```diff
@@ -512,11 +521,21 @@ static int _redisContextConnectTcp(redisContext *c, const char *addr, int port,
...
-addrretry:
-        if ((s = socket(p->ai_family,p->ai_socktype,p->ai_protocol)) == REDIS_INVALID_FD)
+addrretry: {
+        int sock_type = p->ai_socktype;
+
+#ifdef SOCK_CLOEXEC
+        if (c->flags & REDIS_OPT_SET_SOCK_CLOEXEC) {
+            sock_type |= SOCK_CLOEXEC;
+        }
+#endif /* SOCK_CLOEXEC */
```
michael-grunder, line 531:
```suggestion
#endif
```

michael-grunder (changes requested):
Just those two minor changes and then it looks good to me.

collinfunk (approved):

collinfunk:
Looks good to me, thanks!

michael-grunder (approved):
LGTM.

Gave it a local spin and it works as expected. Thanks!

PR #1305, opened 2025-08-04
SWD-2735: update hiredis to v1.3.0

dooglio:


dooglio:
I sometimes hate github.

Issue #1306, opened 2025-09-01
hiredis included through FetchContent should not include CPack


koenvandesande:
When you include hiredis through FetchContent, it causes a warning when the main project also uses CPack:

```
CMake Warning at /usr/local/lib/python3.12/dist-packages/cmake/data/share/cmake-3.31/Modules/CPack.cmake:510 (message):
  CPack.cmake has already been included!!
Call Stack (most recent call first):
  cmake-build-debug-wsl/_deps/hiredis/CMakeLists.txt:98 (include)
```

Could you prevent this by guarding the inclusion of CPack with a check on whether it is the top-level project?
```
if(PROJECT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
  include(CPack)
endif()
```

If needed I could turn this into a PR if that is easier for you.

michael-grunder:
Fixed by #1307

PR #1307, opened 2025-09-09
Issue 1306 fetch content includes cpack

koenvandesande:
This would resolve the issue #1306 for me.


michael-grunder:
If you rebase against the current HEAD i think your tests will pass.

Then we can get them merged.

koenvandesande:
Rebased

PR #1308, opened 2025-09-15
Fix CI for macOS and docker based runs.

michael-grunder:
It used to be neccessary to specify `redis@8.2` to get the latest version but this is now the default.

The only named version is the legacy `redis@6.2`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -97,8 +97,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          brew install openssl redis@8.0
-          brew link redis@8.0 --force
+          brew install openssl redis
+          brew link redis --force
 
       - name: Build hiredis
         run: USE_SSL=1 make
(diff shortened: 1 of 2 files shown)


michael-grunder:
The cross-compilation test failures are unrelated. I'll take a look at those too.

michael-grunder:
OK CI is green now. Rather than add a lot of logic around the docker permission issue when deleting our temp path, i just made it a non-fatal error with a warning.

collinfunk:
Looks reasonable to me. On my machine, `brew install redis` installs Redis 8.2.1.

Though `redis@8.2` still exists as an alias [1], so I don't think the current CI is broken, unless I am missing something.

If we always want the latest version, then this saves us from having to update it.

[1] https://github.com/Homebrew/homebrew-core/blob/main/Aliases/redis%408.2

michael-grunder:
I think it's safe to merge. If there are any issues let me know.

PR #1327, opened 2026-03-20
fix: check for NULL after sdscatlen in sdscatrepr to prevent OOM crash

ghost:
## Summary

Adds NULL checks in `sdscatrepr()` after calls to `sdscatlen()` and `sdscatprintf()`, which can return NULL on allocation failure.

## Bug

`sdscatrepr()` calls `sdscatlen`/`sdscatprintf` in a loop without checking return values. If any allocation fails (OOM), the returned NULL is passed to the next call which dereferences it via `sdslen()`, causing a crash.

## Fix

Add two NULL checks:
1. After the initial `sdscatlen(s, "\"", 1)` before entering the loop
2. After each switch case inside the loop before advancing `p`

Fixes #1318

---

> [!NOTE]
> **Low Risk**
> Low risk: small, localized error-handling change that only affects behavior under out-of-memory conditions and should prevent a NULL dereference crash.
> 
> **Overview**
> Prevents an out-of-memory crash in `sdscatrepr()` by **checking for NULL** after the initial `sdscatlen()` and after each append operation in the loop, returning NULL immediately if an allocation fails.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0f5dbd245b37079054d0a507b1601a6ba0bd7a27. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/sds.c
+++ b/sds.c
@@ -874,6 +874,7 @@ void sdsfreesplitres(sds *tokens, int count) {
  * references must be substituted with the new pointer returned by the call. */
 sds sdscatrepr(sds s, const char *p, size_t len) {
     s = sdscatlen(s,"\"",1);
+    if (s == NULL) return NULL;
     while(len--) {
         switch(*p) {
         case '\\':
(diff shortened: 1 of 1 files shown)


michael-grunder:
I don't think hiredis actually uses `sdscatrepr`.

ghost:
You're right, hiredis doesn't call sdscatrepr anywhere outside of sds.c's own test. The function is only defined in the bundled sds library. I can close this if it's not worth fixing dead code paths.

PR #1335, opened 2026-08-04
Add a `maxdepth` `redisReader` configuration member.

michael-grunder:
Currently hiredis will attempt to parse arbitrarily nested `RESP` replies. During normal operation this is not a problem but a corrupted or maliciously crafted payload could cause a stack overflow.

This commit just adds `redisReader->maxdepth` which has the same semantics as our existing `redisReader->maxelements` so by default hiredis clients will error out if the depth reaches 1024.

This should be far higher than real-world workloads are likely to hit while being much lower than a depth that could overflow the stack.

Vulnurability discovered by:
He Huang, hehuang@swin.edu.au, Swinburne University of Technology, Melbourne, Australia; discovery using NexusSan

---

> [!NOTE]
> **Medium Risk**
> Changes protocol parsing on untrusted input (security-relevant), but defaults are permissive for normal use and behavior matches the existing `maxelements` knob.
> 
> **Overview**
> Adds **`redisReader->maxdepth`** (default **1024** via `REDIS_READER_MAX_REPLY_DEPTH`) so the reply parser rejects deeply nested aggregate replies with a protocol error (**"Max nesting depth exceeded"**) instead of recursing without bound on malicious or corrupted payloads.
> 
> The limit is enforced in **`processAggregateItem`** when entering nested arrays/maps/sets (same pattern as **`maxelements`**). Callers can raise the cap or set **`maxdepth = 0`** for unlimited nesting. README documents **`context->reader->maxdepth`**, and **`test.c`** covers default, at-limit, over-limit, and disabled behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ab1790cabe9122f26914be39b1856fce082bd2b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -638,6 +638,16 @@ unlimited with:
 context->reader->maxelements = 0;
 ```
 
+### Reader max nesting depth
+
+By default the hiredis reply parser limits nested aggregate replies to a depth
+of `REDIS_READER_MAX_REPLY_DEPTH` (currently 1024). If you need to process
+replies nested more deeply, you can increase the value or set it to zero,
+meaning unlimited with:
+```c
+context->reader->maxdepth = 0;
+```
+
 ## SSL/TLS Support
 
 ### Building
--- a/read.c
+++ b/read.c
@@ -557,6 +557,12 @@ static int processAggregateItem(redisReader *r) {
     long long elements;
     int root = 0, len;
 
+    if (r->maxdepth > 0 && r->ridx >= r->maxdepth) {
+        __redisReaderSetError(r,REDIS_ERR_PROTOCOL,
+                "Max nesting depth exceeded");
+        return REDIS_ERR;
+    }
+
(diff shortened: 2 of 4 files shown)


michael-grunder:
Note, this version of the fix would change the API (new `maxelements` member of `redisReader`).

I also have this version which would not require recompilation or `SONAME` bump:
https://github.com/redis/hiredis/compare/master...michael-grunder:hiredis:fix/arbitrary-nesting-depth.v1?expand=1f

I'll let you guys pick which one is preferred.

michael-grunder:
Good call moving the member to the end

uglide:
@michael-grunder GH CI is finicky now, but I've backported your other change to v1.1-v1.4 versions and released new versions.

uglide (approved):

PR #1367, opened 2026-08-31
Add a fuzz target for the RESP reply reader

HuzaifaAbdulRehman:
The reply reader parses bytes the server controls, and it is where
CVE-2021-32765 lived, but it has never been fuzzed. The only harness this
repo carried drove `redisFormatCommand`, and #1295 removed it for good
reason: fuzzing a variadic function with no arguments violates its
contract, so the defects it reported were not defects.

Feeding `redisReaderFeed` has no such problem. Every input is a byte
sequence some server could send.

Two details worth reviewing. The first input byte selects a fragment size,
so replies arrive split across feeds and the incremental parser is
exercised, not just the case where a whole reply lands in one read. And
`maxelements` is capped at 1024, so a header like `*4294967295` reports a
protocol error rather than an out-of-memory that says nothing about the
parser.

Compiles clean under the Makefile's `-Werror -Wall -Wextra -pedantic`.
Ran 1,181,040 executions under libFuzzer and ASan at 7,821 exec/s with no
crashes, growing a 9-file seed corpus to 874.

`projects/hiredis/build.sh` in oss-fuzz still compiles the file #1295
deleted, so those builds have failed since April 2025. I will send the
matching change there once this lands.

---

> [!NOTE]
> **Low Risk**
> Adds fuzz-only code with no changes to the reply parser or library runtime behavior.
> 
> **Overview**
> Adds **`fuzzing/reply_reader_fuzzer.c`**, a libFuzzer harness that drives **`redisReaderFeed`** / **`redisReaderGetReply`** on arbitrary byte streams—the same path server-controlled RESP takes before callers see replies.
> 
> The first input byte picks how large each feed chunk is so parsing is exercised incrementally, not only when a full reply arrives in one call. **`reader->maxelements`** is set to **1024** so huge multi-bulk counts surface as protocol errors instead of fuzzer OOM noise.
> 
> This replaces the gap left after the removed **`redisFormatCommand`** fuzzer and targets the reader area associated with past issues like CVE-2021-32765.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0a8ee08017774d06573914de925aaef7f4b3c8d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


HuzaifaAbdulRehman:
Hi @michael-grunder, could you take a look when you have a chance? I’m happy to make any changes needed.

michael-grunder:
I'm a fan of fuzzing so I don't see a downside.

Right now the way the chunk selection works the fuzzer picks one chunk size and then uses that until it runs out of data

You could use a little RNG helper

```c
/* Marsaglia's xorshift32. The state must be nonzero. */
static uint32_t fuzzRandom(uint32_t *state) {
    uint32_t x = *state;
    x ^= x << 13;
    x ^= x >> 17;
    x ^= x << 5;
    return *state = x;
}
```

HuzaifaAbdulRehman:
Implemented the deterministic xorshift32 chunk sequence and added a small RESP2/RESP3 seed corpus under `fuzzing/corpus/reply_reader`. The update is in `588f686`.

PR #1374, opened 2026-09-17
Drain Linux socket error queue after nonblocking reads

swills:
## Summary

- check `SO_ERROR` after a nonblocking `recv()` returns `EWOULDBLOCK`
- drain Linux extended socket errors that keep `POLLERR`/`EPOLLERR` asserted
- add regressions for the Linux error queue and nonblocking Unix sockets

## Problem

On Linux, an entry in a socket's extended error queue can keep `POLLERR`/`EPOLLERR` asserted after a normal `recv()` returns `EWOULDBLOCK`. Hiredis previously returned a recoverable no-data result without clearing that condition, allowing a level-triggered event loop to spin while the connection remained usable.

Checking `SO_ERROR` alone is insufficient: it can be zero while an extended error-queue entry still asserts error readiness.

## Fix

After `EWOULDBLOCK`, hiredis now reports a nonzero `SO_ERROR` as `REDIS_ERR_IO`. On Linux IP sockets, it then drains `MSG_ERRQUEUE` without blocking before returning the existing no-data result.

The drain is handled in `redisNetRead()` so all event adapters receive consistent behavior. It is limited to 1024 attempts per callback to avoid unbounded work if errors are continuously replenished; exhaustion fails the connection instead of leaving it in a busy loop.

Non-IP sockets, including Unix-domain sockets, skip the Linux-specific drain.

## Testing

The regression uses UDP loopback with `IP_RECVERR` because it produces the relevant Linux error-queue state deterministically and without privileged network setup.

- original implementation fails the regression on Linux amd64 and arm64
- patched implementation passes on both architectures
- Unix-socket regression passes on both architectures
- FreeBSD `gmake check`: `ALL TESTS PASSED (1 skipped)`
- all nine fork CI jobs passed, including Linux with Valgrind, FreeBSD, macOS, Windows/Cygwin, and multiple CMake versions

Related: redis/redis#9956

CI: https://github.com/swills/hiredis/actions/runs/35226833890

---

> [!NOTE]
> **Medium Risk**
> Changes core I/O behavior for all non-blocking reads on Linux IP sockets; mis-handling could surface new errors or affect event-loop latency, though scope is limited to the EWOULDBLOCK path with a bounded drain.
> 
> **Overview**
> Fixes **non-blocking read spin** on Linux when `recv()` returns `EWOULDBLOCK` but the socket still reports error readiness because of a queued extended error.
> 
> **`redisNetRead()`** now treats non-blocking `EWOULDBLOCK` separately from `EINTR`. On that path it reads **`SO_ERROR`** and fails the connection if it is non-zero. On Linux IPv4/IPv6 sockets it then **drains the socket error queue** via non-blocking `recvmsg(..., MSG_ERRQUEUE)` (capped at 1024 entries) before returning the usual “no data yet” `0`, so level-triggered `poll`/`epoll` loops stop re-firing on a healthy idle connection.
> 
> Adds regressions: a Linux UDP/`IP_RECVERR` case that asserts `POLLERR` clears after read, and a Unix socket case that non-blocking read still returns `0` without error when idle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c63297bca15387be32437f8802dfa3c63b8276c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


michael-grunder:
Thanks for the PR

I particularly familiar with Linuxs' extended error queue so read what I could find in the man pages. As far as I can tell the queue is opt-in and AFAIK redis-sentinel doesn't to do that.

So if this is the root cause o sentinel spinning the question Is how is the queue getting enabled in the first place.

As for the PR itself. It adds three additional syscalls per `EWOULDBLOCK` so at the very least would have to be opt-in like our other options.

swills:
> I particularly familiar with Linuxs' extended error queue so read what I could find in the man pages. As far as I can tell the queue is opt-in and AFAIK redis-sentinel doesn't to do that.

You're right. The evidence that led me here was a local `strace` of an affected Sentinel. It repeatedly showed `epoll_wait()` returning `EPOLLERR` for established TCP sockets, followed by `recvfrom()` returning `EAGAIN` and `SO_ERROR` reading as zero.

I inferred that a nonempty error queue was sustaining `EPOLLERR`, but I did not directly observe a queued record or establish how the queue would have been populated. I have now rechecked the Sentinel/hiredis socket setup and found no use of `IP_RECVERR`, `IPV6_RECVERR`, `SO_TIMESTAMPING`, `MSG_ZEROCOPY`, or another identified queue-producing option. The UDP regression demonstrates the drain behavior only for a socket that explicitly enables `IP_RECVERR`; it does not establish the production TCP path.

The PR therefore overstates the root cause, and the additional syscalls on every nonblocking `EWOULDBLOCK` are not justified as unconditional hiredis behavior. I'll close this pending direct evidence of the queued data and its producer. Thanks for catching this.

PR #1375, opened 2026-09-17
Drain Linux socket error queue after nonblocking TCP connect

swills:
## Summary

- drain the Linux socket error queue while a TCP connection is pending and after it completes
- prevent a stale connect-time error from keeping `POLLERR`/`EPOLLERR` asserted
- add a deterministic regression that verifies the connection remains usable

## Problem

Linux can queue an ICMP connection error while a TCP socket is still in SYN state, even when the connection subsequently succeeds. The observed error had `ee_errno == ECONNREFUSED` and remained in `sk_error_queue` after the socket became established.

This does not rely on `IP_RECVERR` being enabled. Linux queued the observed error during the TCP connection attempt, and the regression explicitly verifies that `IP_RECVERR` is disabled.

The stale queue entry keeps `POLLERR`/`EPOLLERR` asserted while `SO_ERROR` is zero and normal `recv()` calls return `EAGAIN`. A level-triggered event loop can therefore dispatch the socket continuously without making progress.

## Fix

During Linux TCP connection establishment, redisCheckConnectDone() now drains MSG_ERRQUEUE before returning either pending or successful completion. Blocking API connections are included because hiredis temporarily uses a nonblocking descriptor while connecting; if such a connection remains pending after an error-queue wakeup, redisContextWaitReady() re-polls using the remaining original timeout. Unix sockets, user-provided file descriptors, and non-Linux platforms retain their existing behavior.

The drain uses `MSG_DONTWAIT`, retries `EINTR`, stops when the queue is empty, and preserves the caller's `errno`. It runs only for library-managed TCP connections. Unix sockets, user-provided file descriptors, and non-Linux platforms retain their existing behavior.

Handling the queue at connect completion clears the stale condition before the async context is marked connected and begins normal command processing.

## Testing

The Linux regression uses software TX timestamping to create the same `sk_error_queue` occupancy condition deterministically, without raw sockets or an ICMP timing race. It then disables timestamping and verifies that:

- `IP_RECVERR`, `SO_TIMESTAMPING`, and `SO_ERROR` are zero
- normal `recv()` returns `EAGAIN`
- `EPOLLERR` remains asserted while the queue entry exists
- successful connect completion clears `EPOLLERR`
- the established connection remains usable and parses `PONG`

The regression failed before the fix and passed afterward.

- Linux amd64: 181 tests passed, 0 skipped
- Linux arm64: 181 tests passed, 0 skipped
- FreeBSD strict C99 compilation and hiredis tests passed
- all nine fork CI jobs passed, including Ubuntu, Rocky Linux with Valgrind, FreeBSD, macOS, Windows/Cygwin, and four CMake versions

Related: redis/redis#9956

CI: https://github.com/swills/hiredis/actions/runs/35277083630

---

> [!NOTE]
> **Medium Risk**
> Changes nonblocking TCP connect completion on Linux, which async clients rely on via `redisCheckConnectDone`, but scope is limited to Linux TCP and other platforms/paths are unchanged.
> 
> **Overview**
> Fixes a Linux-only async networking bug where a **stale socket error queue entry** can leave `EPOLLERR`/`POLLERR` set after a nonblocking TCP connect succeeds, starving level-triggered event loops even though `SO_ERROR` is clear and reads return `EAGAIN`.
> 
> On **Linux TCP** connections, `redisCheckConnectDone()` now drains `MSG_ERRQUEUE` via a new `redisDrainSocketErrorQueue()` helper (non-blocking, `EINTR`-safe, preserves `errno`). Blocking connect waiting in `redisContextWaitReady()` was refactored to **re-poll and re-check completion** until the connect finishes or the timeout elapses, instead of treating a single `poll` wakeup as final.
> 
> Adds a **Linux regression** in `test.c` that uses software TX timestamping to populate the error queue deterministically, then asserts connect completion clears `EPOLLERR` and the socket can still read `PONG`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1083fa3b1e0e202f8afdfdc3177eac0830a8e85d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


swills:
Additional upstream context: The need for this workaround is caused by a Linux kernel bug: IPv4 TCP sockets receive extended-error-queue entries during connection establishment even when hiredis has not enabled `IP_RECVERR`. This behavior was introduced by Linux commit [`45af29ca761c`](https://github.com/torvalds/linux/commit/45af29ca761c275e350cca659856bc56f1035ef9), whose commit message says `IP_RECVERR` should be required, but the IPv4 path omitted that check. Eric Dumazet [proposed adding the missing check in April 2024](https://lists.openwall.net/netdev/2024/04/17/288), but the patch was not merged after [the subsequent discussion](https://lists.openwall.net/netdev/2024/04/18/444). Because the kernel fix was never merged, no upstream Linux release corrects the behavior, so the hiredis workaround cannot be gated on a fixed kernel version.

michael-grunder:
Wow what a great find!

This sentinel 100% CPU issue has been a pain point for years but no one could ever track down why. Kernel bug sure would do it :smile:

Since there not as of yet a way to gate it based on kernel the only prudent thing would be to merge the fix. Also iit's far less impactful to do this housekeeping on connect than doing it every `recv`.

I will play around the PR over the weekend and merge.

@uglide This may solve a years long redis-sentinel 100% CPU bug.

michael-grunder:
Should we also drain in the `EALREADY/EWOULDBLOCK` branch? I suppose a pending error would keep triggering the event loop. It's not really critical since it would only spin untill the connection succeeds or fails but might be worth doing.

net.c
```diff
@@ -360,15 +377,22 @@ int redisCheckConnectDone(redisContext *c, int *completed) {
...
         *completed = 0;
         return REDIS_OK;
     default:
         return REDIS_ERR;
     }
+
+connected:
+    *completed = 1;
+#ifdef __linux__
+    if (c->connection_type == REDIS_CONN_TCP && !(c->flags & REDIS_BLOCK))
```
michael-grunder:
```suggestion
    if (c->connection_type == REDIS_CONN_TCP)
```

Apparently blocking is not immune.

test.c
```diff
@@ -1082,6 +1092,149 @@ static void test_blocking_connection_errors(void) {
...
+    peer_fd = accept(listener_fd, NULL, NULL);
+    assert(peer_fd != -1);
+
+    epoll_fd = epoll_create1(EPOLL_CLOEXEC);
+    assert(epoll_fd != -1);
+    event.events = EPOLLIN;
+    event.data.fd = context->fd;
+    assert(epoll_ctl(epoll_fd, EPOLL_CTL_ADD, context->fd, &event) == 0);
+    assert(setsockopt(context->fd, SOL_SOCKET, SO_TIMESTAMPING,
+                      &timestamping, sizeof(timestamping)) == 0);
```
michael-grunder:
```suggestion
    count = setsockopt(context->fd, SOL_SOCKET, SO_TIMESTAMPING,
                      &timestamping, sizeof(timestamping));
    if (count == -1 && errno == ENOPROTOOPT) {
        test("Nonblocking connect error queue (SO_TIMESTAMPING unavailable): ");
        test_skipped();
        goto cleanup;
    }
    assert(count == 0);
```

I think this is QEMU not supporting the feature

swills:
> Should we also drain in the `EALREADY/EWOULDBLOCK` branch? I suppose a pending error would keep triggering the event loop. It's not really critical since it would only spin untill the connection succeeds or fails but might be worth doing.

Good point. Added this in 055752d: pending nonblocking TCP connects now drain the error queue before returning completed == 0, preventing the queued error from repeatedly waking the event loop while the handshake continues.

net.c
swills: Agreed that blocking mode is not immune. Hiredis temporarily makes blocking TCP connections nonblocking during `connect()`, so I removed the `REDIS_BLOCK` exclusion in `975fdf0`.

I made one additional change rather than applying the suggestion alone: when an error-queue wakeup occurs while a blocking connect is still pending, `redisContextWaitReady()` previously treated `completed == 0` as failure. `SO_ERROR` is zero after the drain, so `redisCheckSocketError()` would substitute the preserved `EALREADY`/`EWOULDBLOCK`. It now re-polls using the remaining original timeout instead. The regression also verifies that completion with `REDIS_BLOCK` clears `EPOLLERR`.

test.c
swills: Agreed. Added the `ENOPROTOOPT` capability skip in `1083fa3`, with a shared cleanup path; other `setsockopt()` errors still fail the test. I verified both the supported path and a forced `ENOPROTOOPT` path on native amd64 and arm64 Linux.

michael-grunder (approved):
LGTM. I'll give the Redis guys time to weigh in but don't see any blockers myself.

michael-grunder:
I see there is a subsequent fix we need too? Unless anyone has objections I'll merge this one and then go through the second PR.

sundb:
@swills Because we need to pick this fix to old version, can you pick this PR to redis?

